PR #22, opened 2021-06-20
Add hybrid-translation.md

cefoo:
I think this file may be lost, as I don't see it in the repo.


bittlingmayer:
It's there, I moved it under /workflows/:

https://www.machinetranslate.org/workflows/hybrid-translation

PR #119, opened 2021-12-13
Update companies.md

luismond:
Change "defence" to "defense"


bittlingmayer (approved):

bittlingmayer:
By the way @cefoo we should think about *finding* all the non-US spellings and also checking new PRs going forward.

https://github.com/marketplace?type=actions&query=languagetool
https://github.com/marketplace?type=actions&query=spelling
https://github.com/marketplace?type=actions&query=spellcheck

PR #122, opened 2021-12-17
Update statistical.md

luismond:
Add a recent method for enhancing SMT: paraphrase-based translation. Source: https://arxiv.org/abs/2109.15119


bittlingmayer:
I'd say that paraphrasing, or any other data augmentation I can think of, is doable and common with neural too.

So probably this belongs here: https://www.machinetranslate.org/customization/training-data

luismond:
I agree. Data augmentation should go into a sub-section of training data.

bittlingmayer:
So is the plan to modify this PR accordingly or open a new one?

luismond:
We will open a new PR

PR #123, opened 2021-12-21
Update parallel-data.md

luismond:
Create "parallel data creation" subsection


bittlingmayer:
I would suggest something like this:
```
### Parallel data creation

Parallel datasets can be created manually or automatically, or created synthetically from monolingual data.
- Human translation or human [post-editing](/workflows/post-editing)
- [Crawling](/research/crawling), [alignment](/research/alignment) and [filtering](/customization/filtering)
- [Back-translation](/customization/back-translation.md)
- [Back-copying](/customization/back-translation.md)
```

Basically, bitext mining and crawling are too intertwined, they're part of the same process.

luismond:
Is it possible to point out a difference here? As far as I know, a crawling method would look for already translated content on aligned corpora (multilingual websites, for example), whereas a bi-text mining approach would try to align sentences in non-aligned corpora.

PR #127, opened 2021-12-23
Create back-translation.md

luismond:
Adding a draft for back-translation.


zouharvi:
Back-translation is also a technical term used in MT model training. Maybe it should be disambiguated here?

I wrote a few papers on backtranslation. The conclusion of them is that it objectively does not increase the final translation quality but it boosts the user confidence. [Backtranslation Feedback Improves User Confidence in MT, Not Quality](https://aclanthology.org/2021.naacl-main.14/)

bittlingmayer:
@luismond I agree with Vilem that this will be very confusing.

Back-translation refers to a technique for creating parallel data from monolingual target-side data (so it goes in that section).

What human translators call back-translation is important in human translation workflows, it's a use of machine translation.  Can you or @cefoo tell us if there are other names for it?  (I usually call it "round-trip translation", but that also has some different implications.)

customization/parallel-data.md
```diff
@@ -12,7 +12,13 @@ description: Parallel data for machine translation
...
 Parallel datasets can include translations for one or more language pairs, and be directioned or directionless.
 
-Parallel data can be created by crawling and aligned monolingual test, and by [back-translation](/customization/back-translation.md) or [back-copying](/customization/back-translation.md).
+### Parallel data creation
+
+Parallel datasets can be created manually, automatically, or created synthetically from monolingual data.
+- Human translation or human [post-editing](/workflows/post-editing)
+- [Crawling](/research/crawling), [alignment](/research/alignment) and [filtering](/customization/filtering)
+- [Back-translation](/customization/back-translation.md)
+- [Back-copying](/customization/back-translation.md)
```
bittlingmayer, line 21: @cefoo Maybe Filtering and Back-translation belong in something called datasets/, not customization/?

cefoo:
> 

In the human translation world, I've seen it referred to as "backtranslation" or "reconciliation". However, "reconciliation" is also final step in an LQA process, and so it can be confusing.

customization/parallel-data.md
cefoo: We have a [Building](https://github.com/machinetranslate/machinetranslate.org/blob/master/ROADMAP.md#building) section. Do you think they can fit there?

zouharvi:
Maybe it could be useful to mention OPUS? https://opus.nlpl.eu/

Some of the links already point to it and it's an amazing library of paralel data between lots of languages.

luismond:
Yes, it seems we are pointing to some sub-corpora. I think we could add OPUS and expand the section.

zouharvi:
OPUS is not a corpus in itself but rather a website that lets you centrally filter among many corpora. It still makes sense to point to those individually (if they are of particular interest).

customization/parallel-data.md
bittlingmayer: It's another option.

PR #128, opened 2021-12-26
add Czech and German languages to the roadmap

zouharvi:
The Czech languages is very well documented in the current research (maybe even more than Russian, see table in WMT [1]). This is also similar for German. Therefore they should probably have separate entries in the wiki.

[1] [statmt.org/wmt21/translation-task.html](https://statmt.org/wmt21/translation-task.html)

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -101,6 +101,8 @@
 - [Japanese](https://github.com/machinetranslate/machinetranslate.org/issues/99)
 - [Indic languages](https://github.com/machinetranslate/machinetranslate.org/issues/100)
 - Russian
+- Czech
+- German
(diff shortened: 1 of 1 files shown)


bittlingmayer:
They'll definitely be included, and if someone contributes an article on a language, I think we should welcome that - but we don't want to list out 100+ languages in the roadmap.

Research is not necessarily the metric here.  Other possible metrics are:
- users
- market size
- products/systems
- datasets
- unique considerations (e.g. Thai)
- world cultural heritage (e.g. Latin, Aramaic)
- Machine Translate reader demand (e.g. search traffic?)
Obviously these are correlated.  And sometimes it makes sense to consider groups of languages.

bittlingmayer (approved):

PR #131, opened 2021-12-29
First draft for String

zouharvi:
First draft for the String article #110.

I am not certain whether encoding should take that much space or whether it should maybe a bit more technical to convey the issue of mismatching encoding.

I did not distringuish between a codetable and encoding (i.e. Unicode vs UTF8/16/32) nor how much space they take up nor endianity.


concepts/string.md
```diff
@@ -0,0 +1,36 @@
+---
+description: String
+---
+
+# String
+
+A string is an object in memory that stores a sequence of characters and colloquially it is the representation of text in the computer.
```
bittlingmayer, line 7: I'm thinking something more like this:

> A string is a sequence of characters.  In computer programming, strings are used to represent text.

concepts/string.md
```diff
@@ -0,0 +1,36 @@
...
+Examples:
+1. `bicycle`
+2. `A man with a telescope 🔭 from the city of Třebíč`
+3. ```
+   Introduction: Lorem Ipsum is simply dummy text of the printing and typesetting industry.
+   
+   Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. 
+   ```
+
+The first example is simply a 7-letter word of English letters.
```
bittlingmayer, line 17: Let's avoid this type of long-term dependency - pointers are hard to read and hard to maintain.

It's a lot of examples and text to make a few basic points.

> A string can be short or even empty.  A string with no characters is the *empty string*.
> > `a`
>
> A string can be long or even infinite.
> `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. ...`
> 
> A string can contain any character, including control characters like *newline*.
> ```
> Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
>
> Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
> ```
> 
> By default, a string is *plain text*.  It does not represent rich formatting or styles, like font-size and color.

concepts/string.md
```diff
@@ -0,0 +1,36 @@
...
+The second example shows that a string can contain spaces, letters from non-English alphabets and even emojis.
+The last example demonstrates that even paragraphs with special whitespace characters, such as the newline, usually written as `\n` are stored as a string.
+
+## Encoding
+
+A particular focus is paid to how the characters are encoded in the string.
```
bittlingmayer, line 23: I'd keep this all simpler and less detailed, and order the flow differently.

> In software, every character has a unique identifier.
> 
> > [ASCII](https://en.wikipedia.org/wiki/ASCII) codes
> > | Character      | Decimal representation |
> > | ----------- | ----------- |
> > | ...      | ...       |
> > | `A`      | 65       |
> > | `B`      | 66        |
> > | `C`      | 67        |
> > |  ...     |  ...        |
>
> ASCII, an early encoding standard, only has 128 characters.  It includes the digits, the letters of the Latin alphabet, common punctuation and control characters.  It does not include `β`, `喂`, or even `é`.
>
> { describe the problems with multiple languages and conflicting standards here }
>
>
> **UTF-8** ...
>
> Other types of encoding:
> - **HTML encoding**
> - **JSON encoding**
> 
> ## Escaping
> ...

concepts/string.md
zouharvi: Updated to your suggestion.

concepts/string.md
zouharvi: > A string can be long or even infinite
:-)

Updated the examples.

concepts/string.md
zouharvi: > Other types of encoding:
>    HTML encoding
>    JSON encoding
This is something different. HTML & JSON encoding are on top of character representation which is handled by the system/standard library.

I am honestly not sure how describe escaping intuitively and well. Maybe you have some suggestions?

Nevertheless I restructured this part.

zouharvi:
@bittlingmayer Incorporated your suggestions.

concepts/string.md
cefoo: Do you think it is perhaps worthy to compare the meaning of the term "string" to that of "segment"?
For our translator audience, it's a clear concept that we come across daily.

@zouharvi @bittlingmayer

bittlingmayer: Probably you should add that, not Vilem, since we're not totally sure what translation industry people mean by a segment.

concepts/string.md
bittlingmayer: Definitely HTML encoding etc are different, but in practice since it's less low-level, people encounter it more in their work.

Probably we need to make a separate article on encoding, and another on escaping.

I don't want to delay your PR with that.

bittlingmayer: I meant to replace

> A particular focus is paid to how the characters are encoded in the string.

with

> In software, every character has a unique identifier.

Let's keep in mind that this article is for people who don't know what a string is!

(Here, you're using "encoded" to explain what encoding is.)

And then the point of the ASCII table is just to give an example of giving a unique id to every character.

concepts/string.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Other encodings
+
+The ASCII standard unfortunately covers only 128 characters which is not sufficient for modern usage across multiple languages many with their own alphabet versions.
+Multiple solutions were proposed that extended ASCII, the most popular ones are Unicode and Windows encoding.
+Among researchers, the UTF8 (a technical specification of Unicode) is the most popular though Windows-based user computers use the later encoding.
+
+The biggest issue is when the text is encoded in one way but the computer tries to interpret it as if it was stored with a different encoding.
+The resulting text is then filled with seemingly-random letters.
+If the third example was stored as UTF8 and then interpreted with a specific Windows encoding, it would look as `A man with a telescope ð­ from the city of TÅebÃ­Ä`.
```
bittlingmayer, line 51: No pointers - here "third example" refers to something somewhere else that could change.

concepts/string.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Encoding
+
+A particular focus is paid to how the characters are encoded in the string.
+The computer does not know what letters are, only numbers.
+Therefore every character needs to have some number assigned (codepoint).
+
+### ASCII
+
+The ASCII standard is a widely popular codetable that maps English and other particularly useful characters to numbers.
```
bittlingmayer, line 31: As a general rule, let's avoid words like "popular", because they can go out of date, although in this case it probably will be true for a long time.

Here "widely" and "particularly" can be removed.

Try not to stuff too much information into one sentence, and avoid ambiguous parses.

> The ASCII standard maps English letters and other common characters to numbers.

bittlingmayer:
Thanks for the updates

I sent a PR now, easier than comments: https://github.com/zouharvi/machinetranslate.org/pull/1/files

I'd be fine if you made *Encoding* its own article.

zouharvi:
Hi @bittlingmayer, thanks for going through this PR and making the updates (merged them to this branch).

bittlingmayer (approved):
Sorry for being late on this

zouharvi:
@bittlingmayer that's ok. I resolved one conflict in the summary file (added a new tab for fundamental concepts) and it's ready for merge.

PR #138, opened 2022-01-03
Update parallel-data.md

luismond:
Update parallel data with Adam's suggestions


bittlingmayer:
1. Links should use `.md`.

2. Those with `/customization/` can be removed

(Over the weekend, we changed how we do things with regard to links.  It should all just work, both on GitHub and on the website.)

PR #176, opened 2022-01-19
Reviewed data-confidentiality and products in the industry section

randyscansani:
I reviewed the articles on products and data confidentiality in the Industry section. In the Product page I changed quite a lot. Hope it's ok!

### DIFF
--- a/industry/products.md
+++ b/industry/products.md
@@ -4,7 +4,7 @@ title: Products
 description: Machine translation products
 ---
 
-Machine translation products include ready-to-use machine translation APIs and applications. Systems can also be made available via integrations.
+Machine translation products include ready-to-use machine translation APIs and applications.  Systems can also be made available via integrations.
 
 They vary by:
 
@@ -12,18 +12,19 @@ They vary by:
 - accuracy
 - price
 - customisation options
-- data confidentiality
-
+- [data confidentiality](data-confidentiality.md)
 and other features, as well as underlying architecture.
 
+Some providers offer ready-to-use solutions that can be integrated immediately.  These usually support many language pairs and their infrastructure is scalable.  An alternative is to train a model from scratch — or to use a pre-trained model — and deploy and maintain it.
 
-Some of the most popular MT providers offer off-the-shelf solutions, i.e. trained MT models that are ready-to-use and can be integrated immediately. An alternative is to train a model from scratch, deploy it and maintain it.
+Unlike raw models, ready-to-use solution usually include:
+- language identification
+- segmentation
+- bridging (pivoting)
+- [glossary](customisation/glossaries)
+- encoding options
+- HTML/XML tag handling
 
-The main advantage of ready-to-use systems is that they are a one-size-fits-all solution, for example they can be used for any text domain. When used in an application, text segmentation and language identification are usually provided. On the other hand, a model trained from scratch might yield better performance on a specific domain if trained on a large amount of high-quality domain-specific bilingual data. However, training and maintaining such models can be a complex task and it requires specific competences.
-Some machine translation providers also offer large pre-trained models that can be fine-tuned. A common application scenario is one where a  pre-trained model is fine-tuned with a relatively small domain-specific data set to achieve high performance on specific contents.
-
-
-Providers offer their machine translation services via a platform, an API or both. Some providers offer limited services for free or paid full services.
 
 | Company | Product |
 | ----------- | ----------- |
--- a/industry/products.md
+++ b/industry/products.md
@@ -15,13 +15,13 @@ They vary by:
 - [data confidentiality](data-confidentiality.md)
 and other features, as well as underlying architecture.
 
-Some providers offer ready-to-use solutions that can be integrated immediately.  These usually support many language pairs and their infrastructure is scalable.  An alternative is to train a model from scratch — or to use a pre-trained model — and deploy and maintain it.
+Some providers offer ready-to-use systems that can be integrated immediately.  These usually support many language pairs and their infrastructure is scalable.  An alternative is to train a model from scratch — or to use a pre-trained model — and deploy and maintain it.
 
-Unlike raw models, ready-to-use solution usually include:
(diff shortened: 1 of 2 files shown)


industry/products.md
```diff
@@ -17,11 +17,13 @@ They vary by:
...
 
 
-An advantage of using a machine translation provider is that the service is already working and can be integrated immediately. An alternative to using a machine translation provider is to train a model from scratch or to use a pre-trained model, and deploy and maintain a machine translation system.
+Some of the most popular MT providers offer off-the-shelf solutions, i.e. trained MT models that are ready-to-use and can be integrated immediately. An alternative is to train a model from scratch, deploy it and maintain it.
 
-Unlike raw models, production machine translation systems usually provide language identification and sentence segmentation.
+The main advantage of ready-to-use systems is that they are a one-size-fits-all solution, for example they can be used for any text domain. When used in an application, text segmentation and language identification are usually provided. On the other hand, a model trained from scratch might yield better performance on a specific domain if trained on a large amount of high-quality domain-specific bilingual data. However, training and maintaining such models can be a complex task and it requires specific competences.
```
bittlingmayer, line 22: I think we made a mistake to ever mention "advantages" - a bit subjective.

We should probably just list out the things that production systems can include, beyond what just the underlying model or research system includes, as a bullet-pointed list:

- language identification
- sentence segmentation
- bridging (pivoting)
- glossary
- encoding options
- HTML/XML tag handling

And most if not all of those should be linked to the relevant article.

And then we should also point out that:
- production systems usually support many languages
- it's on scalable infrastructure

industry/products.md
```diff
@@ -17,11 +17,13 @@ They vary by:
...
+Some machine translation providers also offer large pre-trained models that can be fine-tuned. A common application scenario is one where a  pre-trained model is fine-tuned with a relatively small domain-specific data set to achieve high performance on specific contents.
```
bittlingmayer, line 23: > Some machine translation providers also offer large pre-trained models that can be fine-tuned.

Pre-trained models are also increasingly available for free on the internet.

> A common application scenario is one where a pre-trained model is fine-tuned with a relatively small domain-specific data set to achieve high performance on specific contents.

This isn't really different from the customization in the cloud-based offerings like Google or Microsoft.

industry/products.md
```diff
@@ -17,11 +17,13 @@ They vary by:
 and other features, as well as underlying architecture.
 
 
-An advantage of using a machine translation provider is that the service is already working and can be integrated immediately. An alternative to using a machine translation provider is to train a model from scratch or to use a pre-trained model, and deploy and maintain a machine translation system.
+Some of the most popular MT providers offer off-the-shelf solutions, i.e. trained MT models that are ready-to-use and can be integrated immediately. An alternative is to train a model from scratch, deploy it and maintain it.
```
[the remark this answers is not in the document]
randyscansani: Agreed. The reason why I tried to rephrase this and add some explanation is that I was afraid that the reader might be lost when reading about "MT providers", "raw models", "production systems", "models". Especially "machine translation providers" seems a bit ambiguous to me, since it's often used to refer to the company providing MT solutions, and the company might offer different solutions, not only the ready-to-use ones. My suggestion would be to find a unique term for the ready-to-use solution and one for the model to be trained from scratch so that these can be used consistently. These are just suggestions, of course. If it's just me, we can leave the original!

industry/products.md
randyscansani: Added in the text

industry/products.md
randyscansani: Re-inserted the original wording above

industry/products.md
```diff
@@ -12,16 +12,19 @@ They vary by:
...
-
+- [data confidentiality](data-confidentiality.md)
 and other features, as well as underlying architecture.
 
+Some providers offer ready-to-use solutions that can be integrated immediately.  These usually support many language pairs and their infrastructure is scalable.  An alternative is to train a model from scratch — or to use a pre-trained model — and deploy and maintain it.
 
-An advantage of using a machine translation provider is that the service is already working and can be integrated immediately. An alternative to using a machine translation provider is to train a model from scratch or to use a pre-trained model, and deploy and maintain a machine translation system.
-
-Unlike raw models, production machine translation systems usually provide language identification and sentence segmentation.
+Unlike raw models, ready-to-use solution usually include:
```
bittlingmayer, line 20: Should be plural.

bittlingmayer: What do you think about "usually"?  Is it really true for all the stuff we list below?

Maybe "often"?

I think it's more true/evergreen to say.  "Raw models usually do not include the features included in many production systems:"

industry/products.md
```diff
@@ -12,16 +12,19 @@ They vary by:
...
+- language identification
+- segmentation
+- bridging (pivoting)
```
bittlingmayer, line 23: What would you call this?  I don't really like my wording, that's just what I call it.

industry/products.md
```diff
@@ -12,16 +12,19 @@ They vary by:
...
+- [glossary](customisation/glossaries)
```
bittlingmayer, line 24: Needs a leading `/`

industry/products.md
```diff
@@ -12,16 +12,19 @@ They vary by:
 - accuracy
 - price
 - customisation options
-- data confidentiality
-
+- [data confidentiality](data-confidentiality.md)
 and other features, as well as underlying architecture.
 
+Some providers offer ready-to-use solutions that can be integrated immediately.  These usually support many language pairs and their infrastructure is scalable.  An alternative is to train a model from scratch — or to use a pre-trained model — and deploy and maintain it.
```
bittlingmayer, line 18: I would really prefer "systems" to "solutions" here, just more eng and less enterprisy.  (We have to listen to that stuff enough in our day jobs. ;-))

bittlingmayer: We didn't straight up say what exactly it is, that it's an API or something like that.

bittlingmayer (approved):

PR #187, opened 2022-01-27
Turkish

maidis:
I tried to summarize as best I could. Please feel free to shorten it further.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -53,6 +53,7 @@
 
 * [Languages](languages/languages.md)
   * [Russian](languages/russian.md)
+  * [Turkish](languages/turkish.md)
 
 * [Community](community/community.md)
   * [Communities](community/communities.md)
--- a/languages/turkish.md
+++ b/languages/turkish.md
@@ -1,0 +1,41 @@
+---
+parent: Languages
+title: Turkish
+description: Machine translation for Turkish
+---
+
+The first studies on machine translation in **Turkish** started to emerge in the 90s. Most of these studies were experimental studies and have not reached general-purpose use level.
+
+In addition to machine translation systems designed for the English-Turkish language pair, systems have also been developed between Turkic languages ​​such as Uyghur, Turkmen, Azerbaijani and Tatar.
+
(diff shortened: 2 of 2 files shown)


bittlingmayer:
Thanks for contributing!

I don't want to get into small wording issues, but more make sure we're on the same page about what we're trying to convey, and that will set the overall structure.

At a high-level, this article as written has *too much* information and is *too researchy* - both good things in some other contexts, but machinetranslate.org is not meant to be exhaustive or academic.

Have a look at https://machinetranslate.org/russian (and https://machinetranslate.org/style)

1. Less is more
Try to distill it down to the key details.  Break into sections, and use bullet points.  It should be easy for somebody to scan.  Make the intro sentence a decent description/definition - if I can only read 1 sentence about machine translation for Turkish, what should I know?  For example:

> **Turkish** is an **agglutinative** language, like Hungarian and Japanese, and the language in the Turkic language family with the best machine translation support.

Russian was one of the first languages for which machine translation was researched and developed, and remains one of the best supported languages.

2. Focus on production systems
Which products and APIs have launched support for Turkish and when?  For people who actually want to use machine translation in their own work, that is more important than research projects that may or may not have evolved into anything.

3. Write simple
Think Wikipedia, not a research paper.  For example, use full names, not last names.  Imagine that this all will need to be machine translated.

4. Avoid subjectivity
Is the Turkish monolingual corpus or the English-Turkish parallel corpus really too small?  Compared to what?  It's generally in the top 15-30 by the usual criteria, like Wikipedia size, number of web pages, when it was added to the Google or Facebook UI, when it was supported by Google Translate or Facebook translation etc.  Also note that Ural-Altaic (and even Altaic) is not a language family, but more of a sprachbund - the language family is just Turkic.

5. Evergreen
Similar to above, whether there is or isn't enough Turkish data is something that could change.  An exception would be Turkish being the best supported language in the Turkic language family, which is unlikely to change.  Generally what you wrote conforms to this.

6. Language != country
Avoid "in Turkey"

We could essentially follow the Russian article structure (intro, **Support**, **History**), and add a section for the direct systems between Turkish and other Turkic languages (which is indeed notable, since most other stuff runs via English, and it enabled those other languages to get off the ground).

We could also make this article *Turkic languages*.

Let me know what you think!

PR #188, opened 2022-01-30
Rm trailing slashes from domains

bittlingmayer:
Maybe worth adding to style?


cefoo:
Sure! Added to style.

cefoo (approved):

PR #189, opened 2022-01-31
Update events.md

bittlingmayer:


bittlingmayer:
Arguably we should probably use "Catalonia" if we're going to use "Scotland" and "Florida" :-D

events/events.md
```diff
@@ -5,40 +5,36 @@ title: Events
...
-| Date | Event |
-| ---- | ---- |
-| 2022 | WMT22 |
-| 2022 | AAMT 2022 |
-| 2023 | MT Summit 2023 |
-
-
-## Past events
+| November, 2022 | [**WMT22**](wmt22.md) | |
+| 2022 | AAMT 2022 | |
```
cefoo, line 12: Kozo from AAMT mentioned that AAMT 2022 will probably take place in early December.

cefoo (approved):
Great! This format will be easier to maintain as well, because as soon as an event has passed, we can just change its order in the table instead of adding/losing columns.

PR #190, opened 2022-02-03
Update byte-pair-encoding.md

RHuidrom:


bittlingmayer:
To avoid making multiple PRs, you should merge this into [RHuidrom:patch-1](https://github.com/RHuidrom/machinetranslate.org/tree/patch-1).

PR #196, opened 2022-02-06
Implement Event metadata in JSON-LD for SEO purposes

luisbarrueco:
# Description

This adds a very naive implementation to add [structured data for event pages](https://developers.google.com/search/docs/advanced/structured-data/event) using metadata declared in the front matter.

I've modified these attributes in the page for AAMT2021 in order to "document" what information can be exposed.

You can validate whether Google Search can see rich event data by using the [rich results validation tool](https://search.google.com/test/rich-results?utm_campaign=devsite&utm_medium=jsonld&utm_source=event) (compare what you see when pasting the HTML code of any other event page, with what you get from the AAMT2021 page.)

In order to make this work I had to override what the jekyll-seo-tag plugin outputs in the "head.html" include file from the just-the-docs theme. I found it impossible to do that when it was set up as a "remote theme," so I needed to update the Gemfile so it's installed locally. I'm not 100% sure this doesn't break things (perhaps you originally used jekyll remote themes due to some web hosting factor,) but as far as I can tell by testing this locally, things should work OK.

## Type of PR

- Edits: AAMT 2021 (add sample metadata)
- Other: Adds the ability to interpret the `seo` object in the front matter for pages of type Event.

### Checklist:

- [ ] I have made sure that the URL is not already in use.
- [ ] I have cross-linked relevant articles.
- [ ] I have used the UK spelling.
- [ ] I have kept consistency with the structure of sister articles in the same directory.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -112,7 +112,7 @@ defaults:
 
 # Just the Docs
 
-remote_theme: pmarsceill/just-the-docs
+theme: just-the-docs
 
 logo: favicon.ico
 color_scheme: swissred
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,0 +1,99 @@
+{% comment %}
+This file was copied from $(bundle info just-the-docs --path)/_includes/head.html.
+Based off just-the-docs v0.3.3.
+{% endcomment %}
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+
+  {% unless site.plugins contains "jekyll-seo-tag" %}
+    <title>{{ page.title }} - {{ site.title }}</title>
+
+    {% if page.description %}
+      <meta name="Description" content="{{ page.description }}">
+    {% endif %}
+  {% endunless %}
+
+  <link rel="shortcut icon" href="{{ 'favicon.ico' | relative_url }}" type="image/x-icon">
+
+  <link rel="stylesheet" href="{{ '/assets/css/just-the-docs-default.css' | relative_url }}">
+
+  {% if site.ga_tracking != nil %}
+    <script async src="https://www.googletagmanager.com/gtag/js?id={{ site.ga_tracking }}"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+
+      gtag('config', '{{ site.ga_tracking }}'{% unless site.ga_tracking_anonymize_ip == nil %}, { 'anonymize_ip': true }{% endunless %});
+    </script>
+
+  {% endif %}
+
+  {% if site.search_enabled != false %}
+    <script type="text/javascript" src="{{ '/assets/js/vendor/lunr.min.js' | relative_url }}"></script>
+  {% endif %}
+  <script type="text/javascript" src="{{ '/assets/js/just-the-docs.js' | relative_url }}"></script>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+
+  {% if page.seo.type != "Event" -%}
+    {% seo %}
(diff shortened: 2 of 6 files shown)


_config.yml
```diff
@@ -112,7 +112,7 @@ defaults:
 
 # Just the Docs
 
-remote_theme: pmarsceill/just-the-docs
+theme: just-the-docs
```
bittlingmayer, line 115: Somehow it worked for me locally with just `remote_them`, but actually including it in the Gemfile is probably safer (because it locks the version), more friendly to other developers (if you had problems running locally, someone else will too) and less dependent on GitHub.

_includes/head.html
```diff
@@ -0,0 +1,99 @@
...
+  {% endif %}
+
+  {% if site.search_enabled != false %}
+    <script type="text/javascript" src="{{ '/assets/js/vendor/lunr.min.js' | relative_url }}"></script>
+  {% endif %}
+  <script type="text/javascript" src="{{ '/assets/js/just-the-docs.js' | relative_url }}"></script>
+
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+
+  {% if page.seo.type != "Event" -%}
```
bittlingmayer, line 40: Why is there a `-` here?  (And on lines 42 and 95)

I assume you know what you're doing.

_includes/head.html
```diff
@@ -0,0 +1,99 @@
...
+    {% comment %}
+    Once we've removed the default value from the script tag, we output the rest of the content
+    and we apply our own -very naive- interpretation of what an JSON-LD Event looks like.
+    {% endcomment %}
+    {{ seo_text }}
+    <script type="application/ld+json">
+    {
+      "@context": "https://schema.org",
+      "@type": "Event"
+      {% if page.seo.name %},"name": "{{ page.seo.name }}"{% endif %}
```
bittlingmayer, line 68: Maybe we can do 
```
{% if page.seo.name or page.name %},"name": "{{ page.seo.name or page.name }}"{% endif %}
```
etc etc to avoid having to repeat these fields in the front matter of every page?

_includes/head.html
```diff
@@ -0,0 +1,99 @@
...
+        {% if page.seo.location.name %},"name": "{{ page.seo.location.name }}"{% endif %}
+        {% if page.seo.location.type == "PostalAddress" %}
+          {% if page.seo.location.addressCountry %},"addressCountry": "{{ page.seo.location.addressCountry }}"{% endif %}
+          {% if page.seo.location.addressLocality %},"addressLocality": "{{ page.seo.location.addressLocality }}"{% endif %}
+          {% if page.seo.location.addressRegion %},"addressRegion": "{{ page.seo.location.addressRegion }}"{% endif %}
+        {% elsif page.seo.location.type == "VirtualLocation" %}
+          {% if page.seo.location.url %}"url": "{{ page.seo.location.url }}"{% endif %}
+        {% endif %}
+      }{% endif %}
+      {% if page.seo.description %},"description": "{{ page.seo.description }}"
```
bittlingmayer, line 84: That would also help here:
```
      {% if page.seo.description or page.description %},"description": "{{ page.seo.description or page.description }}"{% endif %}
```
though I realise it is not really much cleaner.

bittlingmayer:
Looks good, thanks so much!

I'm wondering if there is a way we can just overwrite the `seo` variable, or what will happen if just create the `<script...</script>` in custom_head.html...

i.e. any way to avoid forking head.html.

bittlingmayer (approved):

_includes/head.html
luisbarrueco: The idea was to prevent this line from echoing a newline character, in order to get a better looking output.
This was the first conditional I added, I didn't know I was going to have a bunch of such lines (and that's why I didn't add it in the other `if`s. 😞 )

PR #219, opened 2022-03-11
wrap links, resolve #217

### LINKED ISSUE #217: Bug: Link text overflows the Featured article box (phone)
**Description:** The link text overflows the Featured article box in the phone browser.

**Steps to reproduce the behavior:**
1. In your phone, open the browser
2. Go to machinetranslate.org
3. Scroll down to 'Featured article'
5. See error

**Expected behavior:** There should be a line break for the text to be contained inside the 'Featured article' box.

**Screenshots [optional]:**
overflow

**Device:**
 - OS: iOS 15.3.1
 - Browser: Safari
 - Version: 15

zouharvi:
# Description

Fixes #217 by adding `whitespace: pre-wrap` to all links. There was already an override in `publications.md` (on this note I think we should strictly avoid custom `<style>` in separate pages unless there's a strict need.

Notes:
- @bittlingmayer I am not sure if the css is automatically generated from sass on merge or whether there is a custom action that needs to be run.
- I have not tested this as I don't have the prod-ready environment set up locally. Despite that I think this is a reasonable style and I wonder why `nowrap` is the default anyway.

### Checklist:

- [x] I have followed the [style guide](http://machinetranslate.org/style). *the style does not cover "coding"*
- [x] I have made sure that the URL is not already in use. *does not apply*
- [x] I have cross-linked relevant articles. *does not apply*
- [x] I have used the UK spelling. *does not apply*
- [x] I have kept consistency with the structure of sister articles in the same directory. *does not apply*

### DIFF
--- a/_sass/custom/custom.scss
+++ b/_sass/custom/custom.scss
@@ -20,6 +20,11 @@
     font-weight: lighter;
   }
 
+  // all links
+  a {
+    white-space: pre-wrap;
+  }
+
   blockquote {
     background-color: $grey-lt-000;
     padding: 0.1em;
--- a/resources/publications.md
+++ b/resources/publications.md
@@ -17,9 +17,6 @@ td:nth-child(1) {
 td:nth-child(2) {
   min-width: 300px;
 }
-.main-content a {
-  white-space: pre-wrap; 
-}
 </style>
 
 


cefoo:
Hi @zouharvi! I've checked this fix locally, and it works perfectly!
I won't merge yet so that Adam can address your questions.
Thanks for fixing this!

bittlingmayer (approved):

bittlingmayer:
The .sass is not generated, so this change should work.

Thanks for doing this, as you saw, I already had to do it for a specific instance, but this is better.

I'll merge now, it'll deploy, and we can see how we like it.

PR #220, opened 2022-03-11
update footer

### LINKED ISSUE #208: Feature: Put the Machine Translate GmbH and the license in the footer
**Request:** Put the Machine Translate GmbH and the license in the page footer

zouharvi:
# Description

Resolves #208 by adding `CC-BY-SA-4.0` to the footer. I did not add *GmbH* because to me the phrase *Machine Translate GmbH is created and edited by contributors like you!* sounds too corporate and bit non-sensical and it would slightly deter me from contributing (I'd ask myself: *Do I need to be part of this GmbH? I just want to contribute to an open source project a bit. What is all this?*). If @cefoo wishes, I may add it (or some other phrasing).

I also removed the vertical whitespace between *Edited on .. by..* and *Edit this article* to make the footer shorter. I like it more but indeed this is opinionated design and if the maintainers wish, I can roll this back.

Note: Not tested locally.

## Type of PR

Change footer template.

### Checklist:

- [x] I have followed the [style guide](http://machinetranslate.org/style). *does not apply*
- [x] I have made sure that the URL is not already in use. *does not apply*
- [x] I have cross-linked relevant articles. *does not apply*
- [x] I have used the UK spelling. *does not apply*
- [x] I have kept consistency with the structure of sister articles in the same directory. *does not apply*

### DIFF
--- a/_includes/footer_custom.html
+++ b/_includes/footer_custom.html
@@ -18,9 +18,7 @@
     </p>
   {% else %}
     <p id="edited" style="display: none;">
-      Edited on <span id="edited-date"></span> by <a id="edited-username"></a>
-    </p>
-    <p>
+      Edited on <span id="edited-date"></span> by <a id="edited-username"></a>.
       <a
         target="_blank"
         href="https://github.com/machinetranslate/machinetranslate.org/blob/master/{{ page.path }}">
@@ -33,12 +31,10 @@
 <center>
   <p>
       Machine Translate is created and edited by contributors like you!
-  </p>
-  <p>
-    <a
-      href="/contributing">
-      Learn more about contributing
-    </a>
+    </p>
+    <p>
+      <a href="/contributing">Learn more about contributing.</a>
+      Licensed under <a target="_blank" href="http://creativecommons.org/licenses/by-sa/4.0/">CC-BY-SA-4.0</a>.
   </p>
 </center>
 


cefoo:
I tested this locally and I like how it looks!

From my perspective, I think that ***Edit this article*** should be at the top of the page (maybe at the footer as well) because there is a chance that people won't scroll to the end of the article if they find what they need on top.

For instance, if a reader wants to subscribe to ´MT-List´, they google how to do it, and they arrive at our [Communities](https://machinetranslate.org/communities) page, chances are they stop scrolling once they've reached the info they need, not click on our home page, and not realize that they are invited to edit the article or any article or contribute in any way to [Machine Translate](htto://machinetranslate.org).

What do you think, @zouharvi?

zouharvi:
> chances are they stop scrolling once they've reached the info they need, not click on our home page

Some chances are that with more content in the header, *contribute to this article*, they stop reading before they reach the actual information.

I'm a minimalist and like as few things on a website as possible and ideally most of them hidden. :slightly_smiling_face:  That said, I'm (luckily for everyone) not a designer so my opinion shouldn't really affect this decision.

bittlingmayer:
It depends what we're optimising for.

I tend to agree with Vilém on this one that everything has a cost, and 99.9% of people are not editors (and should not be), c'est la vie, and the power users will find the edit functionality if we make it reasonably discoverable OR contact them behind the scenes because they're experts on a topic.

bittlingmayer:
In my eyes the main benefit of having stuff about editing at the top would be for the read-only users to see what great people contributed and when - authority and freshness.

PR #222, opened 2022-03-15
Update results of wmt20

liashahnazaryan:
# Description
I've added the winner list of the wmt20 news translation task. However, there are several things I'd like to discuss, such as 

1. Merging vertical cells where the annotation style is repeated. I've put it separately on each line to make the winner list more comprehensive, but it doesn't seem visually appealing to me, which is why I would like to hear your opinion on this as well. 
2. Excluding the English→Inuktitut results from the second table as it only illustrates the results of the News test set while skipping Hansard. I've included this info in brackets, but as I couldn't find the complete results, I find it misleading to put English→Inuktitut in the winner list with only half the information available. 

## Type of PR
Adding news translation task results of wmt20

### Checklist:

### DIFF
--- a/events/wmt20.md
+++ b/events/wmt20.md
@@ -217,3 +217,49 @@ Friday, November 20, 2020
 | | [***The Tatoeba Translation Challenge – Realistic Data Sets for Low Resource and Multilingual MT***](https://aclanthology.org/2020.wmt-1.139.pdf) <br>Jörg Tiedemann |
 | | [***Human-Paraphrased References Improve Neural Machine Translation***](https://aclanthology.org/2020.wmt-1.140.pdf) <br>Markus Freitag, George Foster, David Grangier and Colin Cherry |
 | |	[***Incorporating Terminology Constraints in Automatic Post-Editing***](https://aclanthology.org/2020.wmt-1.141.pdf) <br>David Wan, Chris Kedzie, Faisal Ladhak, Marine Carpuat and Kathleen McKeown |
+
+## Results
+
+#### Full results of the shared tasks: [***Findings of the 2020 Conference on Machine Translation (WMT20)***](https://aclanthology.org/2020.wmt-1.1.pdf)
+
+### News Translation
+
+#### → English
+> *Systems ordered by Direct Assessment score z-score
+
+| Language pair | Annotation style | Average score | Average z-score (standard score) | System |
+| --- | --- | --- | --- | --- |
+| Chinese→English | Segment rating with document context | 77.5 | 0.102 | `VolcTrans` |
+| Czech→English | Segment rating with document context | 78.3 | 0.118 | `CUNI-DocTransformer` |
+| German→English | Segment rating with document context | 82.6 | 0.228 | `VolcTrans` |
+| Inuktitut→English | Segment rating without document context | 73.1 | 0.168 | `NiuTrans` |
+| Japanese→English | Segment rating with document context | 75.1 | 0.184 | `Tohoku-AIP-NTT` |
+| Khmer→English | Segment rating without document context | 69.0 | 0.168 | `Online-B` |
+| Pashto→English | Segment rating without document context | 67.3 | 0.032 | `Online-B` |
+| Polish→English | Segment rating with document context | 77.2 | 0.131 | `SRPOL` |
+| Russian→English | Segment rating with document context | 79.3 | 0.124 | `Online-G` |
+| Tamil→English | Segment rating with document context | 68.7 | 0.203 | `GTCOM` |
+
+#### English →
+> *Systems ordered by Direct Assessment score z-score
+
+| Language pair | Annotation style | Average score | Average z-score (standard score) | System |
+| --- | --- | --- | --- | --- |
+| English→Chinese | Segment rating with document context | 80.6 | 0.568 | `HUMAN-B` |
+| English→Czech | Segment rating with document context | 85.6 | 0.654 | `HUMAN` |
+| English→German | Segment rating with document context | 90.5 | 0.569 | `HUMAN-B` |
+| English→Inuktitut (News only) | Segment rating without document context | 90.5 | 0.574 | `HUMAN` |
+| English→Japanese | Segment rating with document context | 79.7 | 0.576 | `HUMAN` |
+| English→Polish | Segment rating with document context | 88.6 | 0.672 | `HUMAN` |
+| English→Russian | Segment rating with document context | 91.8 | 0.681 | `HUMAN` |
+| English→Tamil | Segment rating with document context | 83.4 | 0.762 | `HUMAN` |
+| English→Khmer | Segment rating without document context | 77.4 | 0.478 | `GTCOM` |
+| English→Pashto | Segment rating without document context | 73.0 | 0.244 | `GTCOM` |
+
+#### French to/from German (segment rating without document context)
+> *Systems ordered by Direct Assessment score z-score
+
+| Language pair | Average score | Average z-score (standard score) | System |
+| --- | --- | --- | --- |
+| German→French | 90.4 | 0.279 | `OPPO` |
+| French→German | 89.8 | 0.334 | `VolcTrans` |


events/wmt20.md
```diff
@@ -217,3 +217,46 @@ Friday, November 20, 2020
...
+| English→German | Segment rating with document context | 90.5 | 0.569 | `HUMAN-B` |
+| English→Inuktitut (News only) | Segment rating without document context | 90.5 | 0.574 | `HUMAN` |
+| English→Japanese | Segment rating with document context | 79.7 | 0.576 | `HUMAN` |
+| English→Polish | Segment rating with document context | 88.6 | 0.672 | `HUMAN` |
+| English→Russian | Segment rating with document context | 91.8 | 0.681 | `HUMAN` |
+| English→Tamil | Segment rating with document context | 83.4 | 0.762 | `HUMAN` |
+| English→Khmer | Segment rating without document context | 77.4 | 0.478 | `GTCOM` |
+| English→Pashto | Segment rating without document context | 73.0 | 0.244 | `GTCOM` |
+
+#### French to/from German (segment rating without document context)
```
cefoo, line 257: Hi @liashahnazaryan!

Maybe to keep consistency with the previous sub-headers, this header should be spelled with the arrow characters as well.

I would personally use `French <> German` because, as translators, we always list our language pairs with the `<>` characters (or just `>`, as in `English > Spanish`).  This may be different for researchers or academia.
Otherwise, we can just use `Into English`, `From English`, and `French into/from German`.

What do you think?

liashahnazaryan: Agree. I first wanted to use the left-right arrow, but it converts to an emoji, so I chose to go ahead with the `from/to` option. But `<>` seems a good substitute for all cases, and I guess it also applies to WMT more than the other options as they also use arrows instead of `from/to`.

cefoo: > it converts to an emoji

Maybe you need to escape it, like this `\>`?

liashahnazaryan: the problem is solved now I guess, you can see the changes in the updated file
I'll now apply the changes to the wmt21 results as well

events/wmt20.md
```diff
@@ -217,3 +217,46 @@ Friday, November 20, 2020
...
+
+## Results
+
+#### Full results of the shared tasks: [***Findings of the 2020 Conference on Machine Translation (WMT20)***](https://aclanthology.org/2020.wmt-1.1.pdf)
+
+### News Translation
+
+#### → English
+
+| Language pair | Annotation style | Average score | Average z-score (standard score) | System |
```
bittlingmayer, line 229: Do they call it "Annotation style"?

I would probably call it "Evaluation type" or "Evaluation schema" or "Labelling schema".

events/wmt20.md
```diff
@@ -217,3 +217,46 @@ Friday, November 20, 2020
...
+| --- | --- | --- | --- | --- |
+| Chinese→English | Segment rating with document context | 77.5 | 0.102 | `VolcTrans` |
```
bittlingmayer, line 231: Generally, if a column has the same value for all rows, it should not be in the table.

events/wmt20.md
```diff
@@ -217,3 +217,46 @@ Friday, November 20, 2020
 | | [***The Tatoeba Translation Challenge – Realistic Data Sets for Low Resource and Multilingual MT***](https://aclanthology.org/2020.wmt-1.139.pdf) <br>Jörg Tiedemann |
 | | [***Human-Paraphrased References Improve Neural Machine Translation***](https://aclanthology.org/2020.wmt-1.140.pdf) <br>Markus Freitag, George Foster, David Grangier and Colin Cherry |
 | |	[***Incorporating Terminology Constraints in Automatic Post-Editing***](https://aclanthology.org/2020.wmt-1.141.pdf) <br>David Wan, Chris Kedzie, Faisal Ladhak, Marine Carpuat and Kathleen McKeown |
+
+## Results
+
+#### Full results of the shared tasks: [***Findings of the 2020 Conference on Machine Translation (WMT20)***](https://aclanthology.org/2020.wmt-1.1.pdf)
+
+### News Translation
```
bittlingmayer, line 225: > ### News translation

events/wmt20.md
liashahnazaryan: There are two types of annotation styles used here  - segment rating with and without document context, so I've put it in a separate column and included it in the table. Otherwise, I can suggest separating the results for each table with short descriptions about the change of annotation style.

events/wmt20.md
liashahnazaryan: It's called "Annotation style" in the findings, but I guess "Evaluation schema" will also do, as it represents human evaluation configurations

liashahnazaryan (commented):
I've made some changes based on your comments, @bittlingmayer 
The main changes concern the evaluation schema. I deleted the annotation style from the table and separated the results for each table with short descriptions about the change of annotation style.

events/wmt20.md
```diff
@@ -217,3 +217,59 @@ Friday, November 20, 2020
...
 | |	[***Incorporating Terminology Constraints in Automatic Post-Editing***](https://aclanthology.org/2020.wmt-1.141.pdf) <br>David Wan, Chris Kedzie, Faisal Ladhak, Marine Carpuat and Kathleen McKeown |
+
+## Results
+
+Full results of the shared tasks: [*Findings of the 2020 Conference on Machine Translation (WMT20)*](https://aclanthology.org/2020.wmt-1.1.pdf)
+
+### News translation
+
+#### → English
+Human evaluation based on segment ratings with document context.
```
bittlingmayer, line 228: This is not a sentence

events/wmt20.md
```diff
@@ -217,3 +217,59 @@ Friday, November 20, 2020
...
+Human evaluation based on segment ratings without document context.
+
+| Language pair | Average score | Average z-score (standard score) | System |
+| --- | --- | --- | --- |
+| Inuktitut → | 73.1 | 0.168 | `NiuTrans` |
+| Khmer → | 69.0 | 0.168 | `Online-B` |
+| Pashto → | 67.3 | 0.032 | `Online-B` |
+
+#### English →
+Human evaluation based on segment ratings with document context.
```
bittlingmayer, line 249: Ditto

events/wmt20.md
```diff
@@ -217,3 +217,59 @@ Friday, November 20, 2020
...
+
+| Language pair | Average score | Average z-score (standard score) | System |
+| --- | --- | --- | --- |
+| → Chinese | 80.6 | 0.568 | `HUMAN-B` |
```
bittlingmayer, line 253: I guess this was a human, not system?  Just for comparison's sake

So we should always show the top performing system.

If we want to show how well humans did, we can add another column or two (for best and worst).

events/wmt20.md
```diff
@@ -217,3 +217,59 @@ Friday, November 20, 2020
...
+Human evaluation based on segment ratings without document context.
+
+| Language pair | Average score | Average z-score (standard score) | System |
+| --- | --- | --- | --- |
+| → Inuktitut (News only) | 90.5 | 0.574 | `HUMAN` |
+| → Khmer | 77.4 | 0.478 | `GTCOM` |
+| → Pashto | 73.0 | 0.244 | `GTCOM` |
+
+#### French ↔︎ German
+Human evaluation based on segment ratings without document context.
```
bittlingmayer, line 270: ditto

events/wmt20.md
```diff
@@ -217,3 +217,59 @@ Friday, November 20, 2020
 | | [***The Tatoeba Translation Challenge – Realistic Data Sets for Low Resource and Multilingual MT***](https://aclanthology.org/2020.wmt-1.139.pdf) <br>Jörg Tiedemann |
 | | [***Human-Paraphrased References Improve Neural Machine Translation***](https://aclanthology.org/2020.wmt-1.140.pdf) <br>Markus Freitag, George Foster, David Grangier and Colin Cherry |
 | |	[***Incorporating Terminology Constraints in Automatic Post-Editing***](https://aclanthology.org/2020.wmt-1.141.pdf) <br>David Wan, Chris Kedzie, Faisal Ladhak, Marine Carpuat and Kathleen McKeown |
+
+## Results
```
bittlingmayer, line 221: We don't really clearly say that these are the "winning"/"best"/"1st-place" systems.

So "Average" etc looks like it could be average score of all systems...

We need to somehow communicate this clearly.

Also making "System" the 2nd column (before its score) would help.

liashahnazaryan: Made changes according to your comments, @bittlingmayer. I guess the results are now easier to comprehend as the "System" column changed places with the columns representing the scores. 
I've also excluded the "Human" results from the list and replaced them with first in the line "System" results.

events/wmt20.md
```diff
@@ -225,7 +225,7 @@ Full results of the shared tasks: [*Findings of the 2020 Conference on Machine T
 ### News translation
 
 #### → English
-The human evaluation was based on segment ratings with document context.
+The human evaluation used reference-based (monolingual) segment ratings with document context.
```
cefoo, line 228: Maybe we can turn this into a title or into two different sentences so that no info is lost? As it is presented right now, there are two categories: with and without document context.

Also, according to the [Style guide](https://machinetranslate.org/style#minimal-content), we should try to avoid parenthesis.

events/wmt20.md
```diff
@@ -225,7 +225,7 @@ Full results of the shared tasks: [*Findings of the 2020 Conference on Machine T
...
 
 | Language pair | System | Average score | Average z-score (standard score) |
```
cefoo, line 230: Could you add an explanation regarding what Average score and Average z-score are, and how they differ from each other? Maybe these explanations should appear in a different page? What do you think?

Also, have these systems won by their Average score or by their Average z-score? Perhaps it's best to clarify. Have you added only the average scores of the winning systems?

Do you think a bold format for the winning systems would make the table clearer?

bittlingmayer: As I understand it, the only systems listed are the winning systems (for each language pair).

So we don't need to bold, because they'd all be bold.

But we do need to say that, so it's clear.

events/wmt20.md
liashahnazaryan: Thanks for the comments, @cefoo. 
So there are basically two types of direct assessment: reference-based which is the same as monolingual and source-based which is the same as bilingual. I can suggest either deleting the `reference-based` and `source-based` phrases or adding one more sentence to explain their meanings, e.g.

> The human evaluation used monolingual segment ratings with document context. 

> The human evaluation used bilingual segment ratings with document context.

or 

> The human evaluation used monolingual segment ratings with document context. In this configuration, the human assessors should rate the system output by comparing it to the human-generated reference without considering the source input.

> The human evaluation used bilingual segment ratings with document context. In this configuration, the human assessors should rate the source input and system output without any reference translation.

events/wmt20.md
liashahnazaryan: As Adam already stated, the results list only the winning systems. Moreover, these are only official results of the conference. Each year, there may be other types of results as well, e.g., WMT 2021 findings also include source-based and contrastive source-based results of the conference. 
But I haven't included those not to make the overview too complicated for the viewers because those results may need more complex explanations equal to those presented in the findings. 

To make it clear, I suggest writing a short description that can follow the `News translation` title. 

Something like this

> ### News translation
> Official results of WMT20 News Translation Task. Winner systems are listed according to the average z-score and average score. 

Or have a separate description for each translation type

> #### → English
> Official results of WMT20 News Translation Task for translation into English. Winner systems are listed according to the average z-score and average score. The human evaluation used monolingual segment ratings with document context. In this configuration, the human assessors should rate the system output by comparing it to the human-generated reference without considering the source input.


As for adding an explanation regarding what Average score and Average z-score are, I suggest having a separate page for several terms used in WMT, as listing those explanations on each year's conference page will make the pages look heavy with text. For example, the explanation regarding average score and z-score is something like this

> Winner systems are listed according to the average z-score and average score. The average z-score shows how many standard deviations below or above the mean a raw score is. To compute it, human assessment scores for translations are first standardized according to each human assessor’s overall mean and standard deviation score. Average z-scores for individual segments belonging to a given system are then computed before the final overall direct assessment score for that system is computed as the average standard of its segment scores. Results are also reported for average scores for systems, computed in the same way but without any score standardization applied.

And I think there are other terms as well that may need some explanation, such as direct assessment, relative ranking (this was used during previous WMTs), segment ratings with and without document context or in a full document, etc.

Let me know what you think.

events/wmt20.md
cefoo: I agree with deleting that info, as I find `monolingual` and `bilingual` clearer.

Somehow, I still find it hard to understand how this introductory sentence is connected to the table.  Since these are the top systems found through different types of human evaluations, maybe we can make this relation clear by rephrasing to something like 

> Results from human evaluation using monolingual segment ratings with document context

Am I correct?

events/wmt20.md
cefoo: > To make it clear, I suggest writing a short description that can follow the News translation title.
> Something like this
> ### News translation
> Official results of WMT20 News Translation Task. Winner systems are listed according to the average z-score and average score.

I would omit "Official results of WMT20 News Translation Task" as that is clear.
Naming them "Winner systems" is helpful to understand what follows too.



> As for adding an explanation regarding what Average score and Average z-score are, I suggest having a separate page for several terms used in WMT

Totally, we shouldn't add repeated info to all WMT pages.

Would you explain these terms in our [WMT](wmt.md) page or create a new terminology page related to WMT shared tasks?

events/wmt20.md
liashahnazaryan: As WMT uses only human evaluation to determine results and present them in the findings, I didn't want to use the word `results` everywhere because we already have `Results` as a section title. But I get your point, so maybe we should add an overview of the section right after the title. Something like

> ## Results
> The section presents the official results of the news translation task determined through human evaluation. Full results of the shared tasks: [*Findings of the 2020 Conference on Machine Translation (WMT20)*](https://aclanthology.org/2020.wmt-1.1.pdf)

events/wmt20.md
liashahnazaryan: I guess we can add the terms to the WMT page, as I think there are only several terms to explain, and there's no need to create a separate page. Let me know what you think, and I'll create a separate PR for the terms.

cefoo: Sure, go ahead!

Please make sure to link the articles in their corresponding key terms so that readers can go back and forth between them.

events/wmt20.md
```diff
@@ -217,3 +217,61 @@ Friday, November 20, 2020
...
+| → Khmer | `GTCOM` | 77.4 | 0.478 |
+| → Pashto | `GTCOM` | 73.0 | 0.244 |
+
+#### French ↔︎ German
+The human evaluation used bilingual segment ratings without document context.
+
+| Language pair | System | Average score | Average z-score |
+| --- | --- | --- | --- |
+| German → French | `OPPO` | 90.4 | 0.279 |
+| French → German | `VolcTrans` | 89.8 | 0.334 |
```
liashahnazaryan, line 277: Several changes according to our discussions, @cefoo. I'll add links once we have the terminology ready.

PR #228, opened 2022-04-06
Publications: architecture, metrics, vocabulary

lena-voita:
### Model architecture
(1) deleted papers that are not widely used in MT, (2) added Luong et al - considered as one of the foundational papers and often mentioned along with Bahdanau et al; generally this is considered general knowledge of the field. Noticed that it's also in the "Attention section", but think that this is a more suitable place for it. Also swapped places the CNN model with Transformer: the latter one is more important (Transformer is essential, CNNs would be good to know) and the timeline is also in this order.

### Metrics
Made categories for metrics, added one-sentence descriptions for each category.
Deleted:
- Och, 2003 - no evaluation metrics proposed
Added:
 - NIST (Doddington, 2002)
 - chrF (Popovic, 2015)
 - BLEURT
 - BertScore
 - MoverScore
P.S. For Embedding-based category, there were a couple more attempts such as e.g. BLEU2VEC, but they didn’t receive any kind of attention - didn’t put them here.

### Open vocabulary
Added:
SentencePiece (Kudo, 2018) - often used in place of BPE

### DIFF
--- a/resources/publications.md
+++ b/resources/publications.md
@@ -78,7 +78,7 @@ td:nth-child(2) {
 
 |     |     |     |
 | --- | --- | --- |
-| 2018 | [***Subword regularization: Improving neural network translation models with multiple subword candidates.***](https://aclanthology.org/P18-1007.pdf) | Taku Kudo |
+| 2018 | [***Subword regularization: Improving neural network translation models with multiple subword candidates***](https://aclanthology.org/P18-1007.pdf) | Taku Kudo |
 | 2016 | [***Neural Machine Translation of Rare Words with Subword Units***](https://arxiv.org/pdf/1508.07909.pdf) | Rico Sennrich, Barry Haddow, Alexandra Birch |
 
 
@@ -102,9 +102,9 @@ There are two types of metrics: **quality evaluation**  and **quality estimation
 
 #### Similarity-based metrics
 
-These metrics evaluate similarity between machine translation and reference translation. This similarity can be based on:
-* n-gram matching
-* embeddings
+These metrics evaluate similarity between machine translation and reference translation. There are two types of this similarity:
+* n-gram matching-based similarity
+* embedding-based similarity
 
 ##### n-gram matching metrics
 
@@ -112,7 +112,7 @@ These metrics evaluate similarity based on hand-crafted features and rules. For 
 
 |     |     |     |
 | --- | --- | --- |
-| 2015 | [***chrF: character n-gram f-score for automatic MT evaluation.***](https://www.cs.cmu.edu/~alavie/METEOR/pdf/meteor-1.5.pdf) |	Maja Popovic ́ |
+| 2015 | [***chrF: character n-gram f-score for automatic MT evaluation***](https://www.cs.cmu.edu/~alavie/METEOR/pdf/meteor-1.5.pdf) |	Maja Popovic  |
 | 2014 | [***METEOR Universal: Language Specific Translation Evaluation for Any Target Language***](https://www.cs.cmu.edu/~alavie/METEOR/pdf/meteor-1.5.pdf) |	Michael Denkowski, Alon Lavie |
 | 2006 | [***A study of Translation Edit Rate with Targeted Human Annotation***](https://www.cs.umd.edu/~snover/pub/amta06/ter_amta.pdf) | Matthew Snover, Bonnie Dorr, Rich Schwartz, Linnea Micciulla, John Makhoul |
 | 2005 | [***METEOR: An Automatic Metric for MT Evaluation with Improved Correlation with Human Judgments***](http://www.cs.cmu.edu/~alavie/METEOR/pdf/Banerjee-Lavie-2005-METEOR.pdf) | Satanjeev Banerjee, Alon Lavie |
@@ -126,7 +126,7 @@ These metrics use various word embeddings as an alternative to n-gram matching f
 
 |     |     |     |
 | --- | --- | --- |
-| 2020 | [***Bertscore: Evaluating text generation with bert***](https://openreview.net/forum?id=SkeHuCVFDr) | Tianyi Zhang, Varsha Kishore, Felix Wu, Kilian Q. Weinberger, and Yoav Artzi |
+| 2020 | [***Bertscore: Evaluating text generation with bert***](https://openreview.net/forum?id=SkeHuCVFDr) | Tianyi Zhang, Varsha Kishore, Felix Wu, Kilian Q. Weinberger, Yoav Artzi |
 | 2019 | [***MoverScore: Text generation evaluating with contextualized embeddings and earth mover distance***](https://aclanthology.org/D19-1053.pdf) | Wei Zhao, Maxime Peyrard, Fei Liu, Yang Gao, Christian M. Meyer, Steffen Eger |
 
 
--- a/resources/publications.md
+++ b/resources/publications.md
@@ -93,28 +93,22 @@ td:nth-child(2) {
 
 ## Metrics
 
-### Learnable metrics
+Machine translation metrics automatically assess quality of the machine translation output.
+There are two types of metrics: **quality evaluation**  and **quality estimation**.   
+* [Quality evaluation](/quality/quality-evaluation.md) metrics rely on human (reference) translation.
+* [Quality estimation](/quality/quality-estimation.md) metrics ***do not*** rely on human (reference) translation. 
 
-Directly optimize correlation with human judgments.
+### Quality evaluation
 
-|     |     |     |
-| --- | --- | --- |
-| 2020 | [***COMET: A Neural Framework for MT Evaluation***](https://aclanthology.org/2020.emnlp-main.213.pdf) | Ricardo Rei, Craig Stewart, Ana C. Farinha, Alon Lavie |
-| 2020 | [***BLEURT: Learning Robust Metrics for Text Generation***](https://aclanthology.org/2020.acl-main.704.pdf) | Thibault Sellam, Dipanjan Das, and Ankur Parikh |
+#### Similarity-based metrics
 
-### Embedding-based metrics
+These metrics evaluate similarity between machine translation and reference translation. This similarity can be based on:
+* n-gram matching
+* embeddings
 
-Use various word embeddings as an alternative to n-gram matching for capturing word semantics similarity.
+##### n-gram matching metrics
 
-|     |     |     |
-| --- | --- | --- |
-| 2020 | [***Bertscore: Evaluating text generation with bert***](https://openreview.net/forum?id=SkeHuCVFDr) | Tianyi Zhang, Varsha Kishore, Felix Wu, Kilian Q. Weinberger, and Yoav Artzi |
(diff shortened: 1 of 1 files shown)


resources/publications.md
```diff
@@ -95,12 +93,35 @@ td:nth-child(2) {
 
 ## Metrics
 
+### Learnable metrics
+
+Directly optimize correlation with human judgments.
```
bittlingmayer, line 98: *optimise* - European spelling

bittlingmayer: One goal of Machine Translate is that all content will be... machine-translated.

So I worry a little bit about the way we describe each task.  We are using imperatives, which will cause trouble.

Can we think a way to avoid this?

e.g.

> For directly optimising ...

or

> To directly optimise ...

(And maybe no full-stop?)

Or make it a full sentence:

> Learnable metrics should **correlate with human judgements**.

resources/publications.md
```diff
@@ -95,12 +93,35 @@ td:nth-child(2) {
...
+
+|     |     |     |
+| --- | --- | --- |
+| 2020 | [***Bertscore: Evaluating text generation with bert***](https://openreview.net/forum?id=SkeHuCVFDr) | Tianyi Zhang, Varsha Kishore, Felix Wu, Kilian Q. Weinberger, and Yoav Artzi |
+| 2019 | [***MoverScore: Text generation evaluating with contextualized embeddings and earth mover distance***](https://aclanthology.org/D19-1053.pdf) | Wei Zhao, Maxime Peyrard, Fei Liu, Yang Gao, Christian M. Meyer, Steffen Eger |
+
+
+### n-gram matching metrics
+
+Using hand-crafted features, estimate MT quality by counting the number and fraction of n-grams that appear simultaneous in a candidate translation hypothesis and one or more human-generated references.
```
bittlingmayer, line 117: > machine translation

We don't use "MT".  Also it's a bit odd that in the other descriptions, machine translation was not mentioned, just implied.  So maybe it can be here too?  e.g.

> n-gram matching metrics use hand-crafted features to estimate quality.

> They require a human reference translation.

> For example, a metric can count the number of words that appear in both the machine translation and the human translation.

> simultaneously

But I would love to make this sentence shorter and simpler

resources/publications.md
```diff
@@ -95,12 +93,35 @@ td:nth-child(2) {
...
+
+Use various word embeddings as an alternative to n-gram matching for capturing word semantics similarity.
+
+|     |     |     |
+| --- | --- | --- |
+| 2020 | [***Bertscore: Evaluating text generation with bert***](https://openreview.net/forum?id=SkeHuCVFDr) | Tianyi Zhang, Varsha Kishore, Felix Wu, Kilian Q. Weinberger, and Yoav Artzi |
+| 2019 | [***MoverScore: Text generation evaluating with contextualized embeddings and earth mover distance***](https://aclanthology.org/D19-1053.pdf) | Wei Zhao, Maxime Peyrard, Fei Liu, Yang Gao, Christian M. Meyer, Steffen Eger |
+
+
+### n-gram matching metrics
```
bittlingmayer, line 115: To me this seems like prescribing the approach.

Isn't the common denominator here mainly that they depend on a reference translation?

bittlingmayer (commented):
On the substance i.e. what to include and what to remove, LGTM!

I have a few comments about style and wording

resources/publications.md
lena-voita: No, I don't think so. From the contents of the section, I thought that here we are talking about Quality evaluation: this means that all the metrics use reference. This is roughly what I had in mind
Screenshot 2022-04-07 at 14 27 17

So what we could do is
1) instead of "Metrics" have separate sections for Quality estimation and Quality evaluation, link to the article on the website explaining the difference (or explain in one sentence and refer to the article for more details).
2) current content goes to Quality Evaluation
3) something (COMET-QE etc) goes to Quality Estimation.

lena-voita: Btw, if no one is working on the Quality pages, I could take over these, research the topic more and add something like the figure above to the corresponding page.

resources/publications.md
lena-voita: Thank you!
I think the last version can be confusing since here all the metrics should correlate with human judgements (see the picture in the comment below).

How about
_"Differently from the metrics that evaluate similarity between machine and human translations (see the following sections), learnable metrics optimise correlation with human judgements directly."_

resources/publications.md
bittlingmayer: Ah, got it, thanks for explaining

Your suggestion for re-organising the sections makes sense to me.

Basically the dichotomy in the picture is good, but it doesn't really come through on the page right now.

So let's go with your suggestion.

And we should be really careful to keep the wording simple.

resources/publications.md
bittlingmayer: It's good but I am paranoid about potentially ambiguous parses ;-)

So we avoid "как в машине, так и в человеческом переводе".

Probably fine (I tested this one, it didn't happen).

resources/publications.md
bittlingmayer: True

This kind of juxtaposition inside one sentence is a bit haram under our "Simple English" style guide.

I think we can break out the standalone clauses, parentheses etc.

> Learnable metrics optimise correlation with human judgements directly.
>
> They look at only the source sentence and translated sentence.  They do not [require a reference translation](#quality-estimation).

resources/publications.md
bittlingmayer: > Btw, if no one is working on the Quality pages, I could take over these, research the topic more and add something like the figure above to the corresponding page.

Sounds good.  If possible, I think the graphic can just be a Markdown list (with indentations, i.e. tree) or table.

(And the quality pages structure or section structure can also reflect it.)

Then it won't really need the graphic.

(Graphics are hard to edit, translate etc.)

lena-voita: > 

Aaa, I see your point about pictures being hard to edit - I was also a bit worried about it. Therefore, I think the pictures definitely should not contain any fast-changing content (e.g., specific metrics in this case). But can they just reflect the categorisation? I.e., in the picture above remove the specific metrics but keep the structure - this is very likely to be static. To make the pictures editable, I could put the .pptx source along with the picture - in this case, anyone would be able to edit and make a new one. Not as easily as with plain markdown, but for the foundational stuff this might be ok. And each of the figures will still have an accompanying text - this can be machine-translated. What do you think?

Unfortunately, I don't know a way of making the structure above with just markdown lists - can one make several branches as on the figure? If one can, can you link me to an example of such a thing?

cefoo: I think one important item we should consider when thinking about images, besides its translation capabilities, is how they will fit on the website. Please note that the article content section has a somewhat fixed width, and anything that may exceed that width would require readers to left-right scroll. This would also affect phone navigability.

I agree with you that copying the image structure with markdown can be a bit complicated. I think that we may break it into different lists? Or perhaps we can style a table, although I'm uncertain if that is possible.

lena-voita: Yes, we definitely need to think how it'll look on mobile devices!
Re width: can't we just set image width in percentage of the screen width rather than a fixed absolute width value? In html, this works well.

cefoo: Hi Lena!
Should we create a new issue regarding this discussion?

lena-voita: Here it is: https://github.com/machinetranslate/machinetranslate.org/issues/231

resources/publications.md
lena-voita: Just in case, will change to "both the machine translation and the human translation."

resources/publications.md
```diff
@@ -95,12 +93,57 @@ td:nth-child(2) {
...
+* embeddings
+
+##### n-gram matching metrics
+
+These metrics evaluate similarity based on hand-crafted features and rules. For example, a metric can count the number and fraction of n-grams that appear in both the machine translation and the human translation.
+
 |     |     |     |
 | --- | --- | --- |
-| 2020 | [***COMET: A Neural Framework for MT Evaluation***](https://aclanthology.org/2020.emnlp-main.213.pdf) | Ricardo Rei, Craig Stewart, Ana C. Farinha, Alon Lavie |
+| 2015 | [***chrF: character n-gram f-score for automatic MT evaluation.***](https://www.cs.cmu.edu/~alavie/METEOR/pdf/meteor-1.5.pdf) |	Maja Popovic ́ |
```
cefoo, line 115: Seems like the Popovic accent is misplaced. From what I'm seeing at [machinetranslate.org](http://machinetranslate.org), we've spelled it without the accent. What do you think?

cefoo: Also there shouldn't be a period in the paper's name

resources/publications.md
```diff
@@ -95,12 +93,57 @@ td:nth-child(2) {
...
+| 2002 | [***Automatic Evaluation of Machine Translation Quality Using N-gram Co-Occurrence Statistics***](https://aclanthology.org/www.mt-archive.info/HLT-2002-Doddington.pdf) | George Doddington |
 | 2002 | [***BLEU: a Method for Automatic Evaluation of Machine Translation***](https://aclanthology.org/P02-1040.pdf) | Kishore Papineni, Salim Roukos, Todd Ward, Wei-Jing Zhu |
+
+##### Embedding-based metrics
+
+These metrics use various word embeddings as an alternative to n-gram matching for capturing word semantics similarity.
+
+|     |     |     |
+| --- | --- | --- |
+| 2020 | [***Bertscore: Evaluating text generation with bert***](https://openreview.net/forum?id=SkeHuCVFDr) | Tianyi Zhang, Varsha Kishore, Felix Wu, Kilian Q. Weinberger, and Yoav Artzi |
```
cefoo, line 129: With multiple authors, we avoid the "and" and list all names with a comma, as you can see [here](https://machinetranslate.org/style#names-list).

resources/publications.md
```diff
@@ -81,6 +78,7 @@ td:nth-child(2) {
 
 |     |     |     |
 | --- | --- | --- |
+| 2018 | [***Subword regularization: Improving neural network translation models with multiple subword candidates.***](https://aclanthology.org/P18-1007.pdf) | Taku Kudo |
```
cefoo, line 81: Hi dear Lena! ✨
There seems to be a period at the end of the paper's name.

resources/publications.md
```diff
@@ -95,12 +93,57 @@ td:nth-child(2) {
...
+Machine translation metrics automatically assess quality of the machine translation output.
+There are two types of metrics: **quality evaluation**  and **quality estimation**.   
+* [Quality evaluation](/quality/quality-evaluation.md) metrics rely on human (reference) translation.
+* [Quality estimation](/quality/quality-estimation.md) metrics ***do not*** rely on human (reference) translation. 
+
+### Quality evaluation
+
+#### Similarity-based metrics
+
+These metrics evaluate similarity between machine translation and reference translation. This similarity can be based on:
```
cefoo, line 105: We've realized that incomplete introductory sentences for lists are not machine translation-friendly.  We've added it to the [style guide](https://machinetranslate.org/style#lists).

Maybe a good option here could be something like this (please fix if my phrasing is incorrect).

> These similarities can be based on two matching patterns:

lena-voita: Dear Cecilia,
Thank you for finding these! Updated.

PR #248, opened 2022-05-12
add personal webpages

zouharvi:
# Description

Related to  #247, this PR adds a few personal links (some were on request). Some people have multiple personal pages and in that case I tried to use the most "evergreen" one, which is usually the one they maintain themselves or that which is listed in their Google Scholar. I skipped multiple people because I don't know much about them and was unsure which to choose.

## Type of PR

- Edits `people/` by adding personal links where available. 

### Checklist:

### DIFF
--- a/people/andy-way.md
+++ b/people/andy-way.md
@@ -6,3 +6,5 @@ layout: coming_soon
 title: Andy Way
 description:
 ---
+
+[Personal webpage](https://www.computing.dcu.ie/~away/).
--- a/people/jaime-carbonell.md
+++ b/people/jaime-carbonell.md
@@ -6,3 +6,5 @@ layout: coming_soon
 title: Jaime Carbonell
 description:
 ---
+
+[Personal webpage](http://www.cs.cmu.edu/~jgc/).
--- a/people/martin-popel.md
+++ b/people/martin-popel.md
@@ -6,3 +6,5 @@ layout: coming_soon
 title: Martin Popel
 description:
 ---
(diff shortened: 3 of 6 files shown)


zouharvi:
This PR also adds Matt Post, a very influential researcher in MT.

Question: Currently the ordering is given by the number in the file header. Isn't there an easier way to e.g. sort them alphabetically?

cefoo:
> Currently the ordering is given by the number in the file header. Isn't there an easier way to e.g. sort them alphabetically?

It's in inverse chronological birth order (may not be exact, I could not figure some exact birthdates out).

I should probably have spaced the ones we have better so that inserting new people is not too burdensome.
I'll work on that.

zouharvi:
> It's in inverse chronological birth order (may not be exact, I could not figure some exact birthdates out).

I didn't know that and added Matt to the end of the list.

cefoo:
Don't worry!

I'll fix the order (and space out the other People articles' nav order) after we merge.

zouharvi:
Maybe this could also be discussed whether sorting by age is the best way for this? I understand that e.g. the "previous generation" should probably be shown at the bottom. But at the same time this doesn't feel entirely right for someone to get the spotlight just because they were born the latest.

cefoo:
Sure, I agree! Should we move this to the discussion tab?

zouharvi:
Alright. Could you merge this PR though? :slightly_smiling_face:

PR #269, opened 2022-06-29
Create translation-management-systems.yaml

### LINKED ISSUE #211: Article: MT support in CAT/TMS
**Article title:** MT integrations in CAT/TMS

**Directory:** Resources

**Description**: Machine translation integrations and possibilities in CAT/TMX systems

**Definition**:
What CAT/TMS support MT?
What MT systems are supported?
Is the MT connection added by default in the configuration or do you need to access via a plugin?
How do they deal with customisation?
etc.

cefoo:
# Description

Fixes #211 

## Type of PR

- Creates the article _translation-management.systems.yml_

### Checklist:

About this PR:

- `custom` means that the tool can access the user's system custom model.
- `glossary` means that the tool supports MT glossary selection from its interface regardless of whether the MT system supports glossary/dictionaries or not.
- Memsource Translate enables customization of certain engines. I have inserted these details in Memsource's entry, in the engine level.
- Although some engines use "terminology", "glossary", or "dictionary", I've unified the terminology to "glossary".
- For `modernmt`, I inserted `custom: true`, but not `adaptive: true` as this feature is not easily accessed from TMSs.


_data/translation-management-systems.yaml
```diff
@@ -0,0 +1,669 @@
+---
+#- id: [Lowercase, words separated by hyphen]
```
bittlingmayer, line 2: This is good.

We should probably use https://json-schema.org/.

Then it can be automatically enforced.

It works fine with YAML, not just JSON.

bittlingmayer (commented):
I think we can remove `description`.

If it contains important information - e.g. whether a TMS has a CAT or not, whether it's cloud-based or local - that should just be in the structured data, if it's not already.

Later, we could always generate a description from structured data.

bittlingmayer:
I think for `machine_translation_integrations` and `machine_translation_plugins`, we should use a dictionary not a list.

So
```
    - amazon
      custom: true
      glossary: true
```
instead of 
```
    - service: amazon
      custom: true
      glossary: true
```

For the top-level items, it's different.  But not for these nested things.

_data/translation-management-systems.yaml
```diff
@@ -0,0 +1,669 @@
...
+#    - name:
+#      description:
+
+- id: crowdin
+  name: crowdin
+  description: Translation management system
+  url: https://www.crowdin.com
+  cat: true
+  documentation: https://support.crowdin.com/configuring-machine-translation-engines/
+  machine_translation_integrations:
```
bittlingmayer, line 32: Crowdin really only supports custom for Google and Microsoft, and tone for DeepL.

For anything else, you need to write your own custom integration app (or use Intento).

Screenshot 2022-07-09 at 10 44 04

_data/translation-management-systems.yaml
```diff
@@ -0,0 +1,669 @@
...
+
+    - service: tilde
+      custom: true
+
+    - service: ubiqus
+      glossary: true
+
+    - service: yandex
+
+  quality_integrations:
```
bittlingmayer, line 256: Is this about QE?  I don't think that their QE feature is called Memsource Translate, if it has a name.

_data/translation-management-systems.yaml
```diff
@@ -0,0 +1,669 @@
+---
+#- id: [Lowercase, words separated by hyphen]
+#  name:
+#  description:
+#  url:
+#  cat: [Boolean]
+#  documentation: [Link]
```
bittlingmayer, line 7: should be `documentation_url`

_data/translation-management-systems.yaml
```diff
@@ -0,0 +1,669 @@
+---
+#- id: [Lowercase, words separated by hyphen]
+#  name:
+#  description:
+#  url:
+#  cat: [Boolean]
```
bittlingmayer, line 6: if this is always true, maybe we don't need it?

_data/translation-management-systems.yaml
```diff
@@ -0,0 +1,669 @@
...
+    - service: systran
+      custom: true
+
+    - service: textra
+
+    - service: tilde
+      custom: true
+      glossary: true
+
+    - service: tmxmall
```
bittlingmayer, line 164: Another aggregator. :-)

bittlingmayer:
We need to make sure that any engine here is listed in engines.yml.

I think we should call it

```
engine_integrations:
```

If we want to keep aggregators separate and out of engine.yml, then we should just do:

```
engine_aggregator_integrations:
```

And later we can add engine_aggregator.yml.

_data/translation-management-systems.yaml
```diff
@@ -0,0 +1,669 @@
...
+
+    - service: watson
+      custom: true
+
+  features:
+    - name: Local machine translation engine
+      description: Users can use their local machine translation in the tool
+
+- id: trados
+  name: Trados Studio
```
bittlingmayer, line 394: This isn't a TMS though.

bittlingmayer: If we want to have CATs like Trados and OmegaT here, then we should rename this file to integrations.yml.

And then we could have fields to say whether it's a CAT, TMS or both, although I don't think it's that important, since that's not our mission, and there are other resources on that.

bittlingmayer:
Features are repetitive enough that we should just make those that are common and important top-level boolean fields.

```
- routing: [ ... ]
- fuzzy_repair: [ ... ]
- ...
```

However, I can live without them.

_data/translation-management-systems.yaml
cefoo: Yes, it is. I should rename it to `quality_estimation`.
I guess I could make `quality_integrations` a boolean and, if the service is named, then add `name`.

_data/translation-management-systems.yaml
cefoo: Yeah, there are several aggregators... Should we separate them into a different category or delete them altogether?

_data/translation-management-systems.yaml
cefoo: Yeah, I think it's important to have Trados or OmegaT here. I'll rename the file.
The thing with Trados is tricky. Many RWS products only support RWS MT, but all other are available through Trados and the RWS AppStore. For instance, when researching tools for Passolo, RWS support always suggests exporting the file for Trados and pretranslating there.

_data/translation-management-systems.yaml
cefoo: I've changed `cat` to `type`, and made it a list, where values are either `cat`, `tms`, or both.
An alternative could be that, instead of having `cat: [boolean]`, we can have `tms: [boolean]` as this is the characteristic that changes.

bittlingmayer:
> Although some engines use "terminology", "glossary", or "dictionary", I've unified the terminology to "glossary".

It's good to start the article with

> ... **glossary** or **terminology** or ** dictionary** is ...

Or can you think of another way to weave in these synonyms?  So people find this page if they search for eg *customize mt with dictionary*.

bittlingmayer:
> 
> Will things be broken if mentioned here but not there. Shall I at least generate empty entries?

No, it's fine.  The important thing is to remember to do this all later.

_data/translation-management-systems.yaml
bittlingmayer: in some ways they're more like integrations than engines

I'd keep them separate.

_data/translation-management-systems.yaml
bittlingmayer: Then it needs to always be a list.

`types: [ ... ]`

But I would lean towards the boolean approach.

_data/translation-management-systems.yaml
bittlingmayer: Boolean has problems.  I'd just call it *Memsource quality estimation*

bittlingmayer: Later we'll make a list of those too.

_data/integrations.yml
```diff
@@ -0,0 +1,626 @@
+---
+#- id: [lowercase, words separated by hyphen]
+#  name:
+#  url:
+#  type: [ tms, cat ]
+#  platform: [ cloud-based or desktop ]
```
bittlingmayer, line 6: Here it's not super clear that this one is not a list but the one above is.

I'd probably use `|` instead of `or` here.

bittlingmayer: I'd go with a schema more like this:

```
# - id: <string> # slug
#   name: <string>
#   url: <string> # link
#   types: [ 'tms' | 'cat', ... ] # Functionality that it has
#   platform: 'cloud-based' | 'desktop'
#   open_source: true | false
#   engine_integrations:
#      - service: <string> # slug of the corresponding engine id
...
```

bittlingmayer, line 6: What if it's on both desktop, mobile and cloud?  Maybe `platforms` (plural)?

That said, I don't think it's our business to track this.

cefoo:
> > Although some engines use "terminology", "glossary", or "dictionary", I've unified the terminology to "glossary".
> 
> It's good to start the article with
> 
> > ... **glossary** or **terminology** or ** dictionary** is ...
> 
> Or can you think of another way to weave in these synonyms? So people find this page if they search for eg _customize mt with dictionary_.

What about [keywords](https://schema.org/keywords) in the article's front matter?

_data/translation-management-systems.yaml
cefoo: I put `memsource-mtqe` so that it's the ID.
I'll add it to `quality.yml` so that refers back to that file.

_data/integrations.yml
```diff
@@ -0,0 +1,632 @@
...
+- id: trados
+  name: Trados Studio
+  url: https://www.trados.com/products/trados-studio
+  type: [ cat ]
+  documentation_url: https://docs.rws.com/980998/261377/trados-studio-2022/welcome-to-the-------------trados-studio-2022-documentation
+  engine_integrations:
+    - language-weaver:
+        custom: true
+
+  engine_plugins:
```
martab0, line 389: Trados can also use Google and Microsoft MTs - they are hidden in this plugin: https://appstore.rws.com/language/app/mt-enhanced-plugin-for-trados-studio/604/

_data/integrations.yml
```diff
@@ -0,0 +1,632 @@
...
+#   quality_estimation:
+#      - <string> # slug
+#        custom: true | false
+#
+#   local_engine_support: [ boolean]
+#   fuzzy_repair: [ boolean ]
+#   fuzzy_repair_plugin: [ link ]
+#   multiple_engine_suggestions: [ boolean ]
+#   routing: [ boolean ]
+
```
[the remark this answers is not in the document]
cefoo: Across is an important enough tool to include in this PR.
However, I believe that it's key that we keep our scope controlled, as there are thousands of TMSs out there and keeping the article updated will become too hard if all are included.

bittlingmayer:
Looking forward, I think fuzzy repair is not worth covering, because:

- With adaptive, MT subsumes that, it won't be the TMS doing it.

- TMSes are not ML-companies, most not even tangentially, so MT functionality will develop faster than equivalent TMS functionality.

- There are very few TMSes offering this and each does it differently, so it's not easy to define fields yet.

- It's not MT anyway, so it's out of scope.

So I'd so don't bother for now, and then we can revisit this later, if it turns out I'm wrong about this.

martab0:
Improving fuzzy matches with MT is available both in Trados (Fuzzy Repair) and in memoQ (Match Patch), however I have never used any of those :) I can ask around some experienced translators if they do and whether they find it useful.

bittlingmayer:
We'll see what happens.  RWS Language Weaver, i.e. the MT product, just launched adaptive.  So it's not clear yet where the fuzzy repair will live in the future - in the MT, or in the TMSes, or in both.

But it'll be easier on contributors if we don't cover features of TMS that beyond their MT API integrations.

PR #292, opened 2022-07-24
Change titles in "Parallel data"

bittlingmayer:
- Remove the main title - it's autogenerated at the top of the article anyway, so now it's repeated
- Remove "parallel data" from titles, it's implied (hence why we don't say "Parallel data challenges")

### DIFF
--- a/customisation/parallel-data.md
+++ b/customisation/parallel-data.md
@@ -6,18 +6,18 @@ description: Parallel data for training machine translation
 featured: true
 ---
 
-**Parallel data** or **parallel corpora** are datasets of **translation pairs** - sentences and their translations.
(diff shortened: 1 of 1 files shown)


bittlingmayer:
@cefoo Should we use "data set" or "dataset"?

May need to add this to the style guide.

cefoo:
> @cefoo Should we use "data set" or "dataset"?
> 
> May need to add this to the style guide, if those we use don't say anything on it.

I've seen both, even under WMT, but a simple Google search shows more results for "dataset".

I think it would be useful to have a `Terminology` section in the SG. I see what else, already included in the SG, can be reorganized under `Terminology`.

PR #295, opened 2022-07-24
Update social icons

bittlingmayer:


bittlingmayer:
This PR is obsolete now, we already refactored this.

PR #323, opened 2022-09-07
Draft fro SLT in Practice, article for MTM22

pyRis:
# Description

Article: SLT in Practice

## Type of PR

- Creates the article(draft) SLT in Practice

### Checklist:

- [ y] I have read the [contributing guidelines](contributing.md).
- [ y] I have followed the [style guide](http://machinetranslate.org/style).


cefoo:
Hi @pyRis!
Thanks a lot for your first draft PR! 🎉

Here are a couple of notes for the second draft:

- Please note that, in order to minimize potential translation problems when we machine translate the site, we use full formats instead of acronyms. Therefore, please expand things like `SLT`, `MT`, `ASR`, etc. You can see more about the style we aim for at our [style](http://machinetranslate.org/style) page.
- Try to use evergreen vocabulary. This way, if there are breakthroughs in the technology and if things change, the article does not stop being true. Please note the difference between saying "Currently, the most used machine translation approach is neural machine translation" and "By 2022, the most used machine translation approach was neural machine translation". If something new emerges, the "Currently, ..." phrasing stops being true, but the "By 2022,..." phrasing doesn't.
- Perhaps it would be best to make the article impersonal. We want Machine Translate to be equally useful for researchers, developers, translators or localization experts, and not all of them will be identified by the use of "we". 😉
- Have you considered writing a short article about what SLT is? :) Very simple and short, you can follow the [neural machine translation](http://machinetranslate.org/neural-machine-translation) article for structure.

Looking forwards to see how the article develops! ☺️

zouharvi:
@pyRis do you have time to finish this and have your name be engraved in the contributors list forever? :slightly_smiling_face:

bittlingmayer:
This seems like a draft, so I changed it to a draft.

PR #334, opened 2022-09-23
Pin pmarsceill/just-the-docs to v0.3.3 to avoid performance degradation

luisbarrueco:
# Description

pmarsceill/just-the-docs recently published a Release Candidate for [v0.4.0.rc2](https://github.com/just-the-docs/just-the-docs/releases/tag/v0.4.0.rc2). It affected the build time for this project, which went from around ten minutes to over an hour.

Screen Shot 2022-09-23 at 18 33 39

This PR pins the remote theme to v0.3.3, which is the version that was in use since the beginning.

## Type of PR

- Other (build time improvement)

### Checklist:

- [ ] I have followed the [style guide](http://machinetranslate.org/style). **N/A**

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -133,7 +133,7 @@ defaults:
 
 # Just the Docs
 
-remote_theme: pmarsceill/just-the-docs
+remote_theme: pmarsceill/just-the-docs@v0.3.3
 
 logo: favicon.ico
 color_scheme: swissred


bittlingmayer:
Thank you, Luis!

pdmosses:
@bittlingmayer Thanks for the test of v0.4.0.rc2, and for the feedback!

Testing a clone of `machinetranslate` locally. I'm getting the following build times (Jekyll v3.9.0, macOS 12.6,  MBA M2):

* v0.3.3: 425 seconds (~ 7 minutes)
* v0.4.0.rc1: 316 seconds  (~ 5 minutes)
* v0.4.0.rc2: 2284 seconds (~ 40 minutes)

I've been trying to optimise the build of your site to get it close to the v0.4.0.rc1 time, but so far without success.

Could you perhaps try pinning to v0.4.0.rc1, to check that you also get a (slightly) lower build time than with 0.3.3?

A site with 140 pages [reported](https://github.com/just-the-docs/just-the-docs/pull/956#issuecomment-1243374685) a different progression:

* v0.3.3: 70 seconds
* v0.4.0.rc1 with simplified nav: 7 seconds
* v0.4.0.rc2: 12 seconds

The navigation changes in v0.4.0.rc2 were motivated by an inefficiency pinpointed by profiling on that site. A site with more than 300 pages [reported](https://github.com/just-the-docs/just-the-docs/issues/863#issuecomment-1242742371) a similar improvement: in v0.4.0.rc1 ~3 minutes , in v0.4.0.rc2 ~30 seconds (using Jekyll 4).

luisbarrueco:
Hi @pdmosses ! Sorry for my delayed reply.
I've tested v0.4.0.rc1 in my machine, and I get a slightly faster build time compared to using v0.3.3 (11m38s vs. 13m42s)

pdmosses:
Hi @luisbarrueco! Thanks for testing the build time with v0.4.0.rc1 – and good that it's a bit faster, rather than slower. The navigation in the forthcoming v0.4.0 is now to be based on that from v0.4.0.rc1 (with some further optimisation), instead of v0.4.0.rc2.

pdmosses:
@luisbarrueco Please see the [profiling results](https://just-the-docs.github.io/just-the-docs-tests/navigation/build-time/) for the final(?) pre-release of v0.4.0. The build time for he `machinetranslate` website isn't quite as good as for v0.4.0.rc1, unfortunately: I've found optimization of Liquid/Jekyll code quite difficult – especially as I'm not familiar with the details of how Jekyll actually works…

The simplest way of significantly reducing the build time for the `machinetranslate` website might be to start using Jekyll 4.2.2. It's possible that the breaking changes between Jekyll 3 and 4 don't affect your site. Unfortunately the `github-pages` gem isn't compatible with Jekyll 4, so it would require updating the `Gemfile` with direct dependence on other gems that are compatible.

pdmosses:
Hi @luisbarrueco and @bittlingmayer,

For a comparison of build times for various websites when using the recent pre-releases and v0.3.3, see the [profiling results](https://just-the-docs.github.io/just-the-docs-tests/navigation/build-time/).

I'm sorry that the time for building the machinetranslate site has increased in v0.4.0.rc3, relative to v0.3.3 and v0.4.0.rc1. It seems that an added test for the number of irrelevant excluded pages (needed for optimising another large site) blocks some optimisations that Jekyll otherwise applied to the subsequent code. I'll try to take another look at it before @mattxwang releases v0.4.0, in case I've overlooked some way of avoiding the slowdown.

One way of reducing your build time significantly would be to use Jekyll 4. But then you'd need to drop the `github-pages` gem, and you wouldn't be able to use the classic default build on GH Pages. It's probably best to wait with that until Just the Docs switches to Jekyll 4 and stops using Jekyll 3 in its own workflow.

PR #335, opened 2022-09-23
Update string.md

mikethetexan:
# Description

## Type of PR

- Edits [concepts/string.md](https://github.com/machinetranslate/machinetranslate.org/compare/master...mikethetexan:machinetranslate.org:patch-1#diff-2ba781180ad813437f32d012ce88e5055ef9e73d73fb2431510991efc6cb1e2c)

### Checklist:

- [X] I have read the [contributing guidelines](contributing.md).
- [X] I have followed the [style guide](http://machinetranslate.org/style).

### DIFF
--- a/concepts/string.md
+++ b/concepts/string.md
@@ -25,12 +25,14 @@ So a string can even be used for a paragraph or book.
 
 ## Encoding
 
-In software, every character has a unique number.
+Computers deal with numbers, not with characters.  In order to represent a character, it has to be encoded into a series of bits.  Since computers deal with bytes (8 bits), a character is typically encoded into one or several bytes of data.
 
-The computer does not know what letters are, only numbers.
-So every character needs to be represented by a unique number (codepoint).
+First, one has to decide how many characters we care to represent.   That's what is called the **character set**.   Each characted is typically assigned a **codepoint**.
+Then, one has to decide how to encode these characters, i.e. which byte values to assign for that codepoint.
 
-For example, the **ASCII** standard maps English and other particularly useful characters to numbers.
+A character set can have one or several possible encodings. 
+
+The **ASCII** standard defines 128 characters and a single encoding.  Here is an example of a few ASCII characters and their decimal byte values:
 
 | Character | Decimal representation |
 | ... | ... |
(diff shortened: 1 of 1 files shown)


concepts/string.md
```diff
@@ -46,13 +48,16 @@ ASCII only has 128 characters.
...
+
+In 1991, a new standard emerged called Unicode.  Its goal was to represent as many characters as possible in a single character set and solve the interoperability issue. Its first version contained over 7000 characters.  As of September 2022, Unicode is in version 15.0 and contains 149,186 characters or codepoints!
+
+Unicode has many different encodings. The most common one is **UTF-8**, but other exist like **UTF-16** or even **UTF-32**.  UTF-8 is a variable length encoding, while UTF-32 is fixed length (as long as the character set doesnt try to represent more than 4 billion chars!). UTF-32 is a memory hog, but it is predictable since it's fixed length and it has the advantage that the byte values are an exact match to the codepoint value.  UTF-16 is also variable length, not fixed length, a mistake often made by newbie developers.  In UTF-16, in order to represent characters beyond the first plane of Unicode (Basic Multilingual plane), the encoding uses **surrogate pairs**.  Don't ever assume that a character is always encoded with 2 bytes in UTF-16!
```
cefoo, line 58: This information is super interesting, but I fear it might not be a little out of scope for Machine Translate, as this article is about strings.
Please bear in mind that we try to [keep content minimal](https://machinetranslate.org/style#machine-translation-ready).
Do you think we can summarize this part into fewer sentences, just to give minimal history?

concepts/string.md
```diff
@@ -46,13 +48,16 @@ ASCII only has 128 characters.
...
 
-The most common problem is when text is encoded with one standard, but decoded with another.
-The result is often unreadable.
+A common problem is that legacy character sets and their encodings are still supported by many operating systems (for compatibility reasons with old software), while newer systems may use a Unicode encoding. When text is encoded with one standard, but decoded with another, the result is often unreadable.
```
cefoo, line 60: I like giving more detail to the problem!
I would just put it into shorter sentences, to avoid potential MT problems in the future.

cefoo (commented):
Hi @mikethetexan!

Thank you so much for your edits and all your hard work!

I left some comments to your edits to try to align the article with our Style Guide and the structure of other articles in the repo.

Please let me know what you think. Let's work together! 😊

Thanks!

concepts/string.md
```diff
@@ -25,12 +25,14 @@ So a string can even be used for a paragraph or book.
 
 ## Encoding
 
-In software, every character has a unique number.
+Computers deal with numbers, not with characters.  In order to represent a character, it has to be encoded into a series of bits.  Since computers deal with bytes (8 bits), a character is typically encoded into one or several bytes of data.
```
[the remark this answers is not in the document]
bittlingmayer: I'd phrase it as:

> At the low level, computers use numbers, not characters.  So each character is actually represented by a unique number.

This part is optional, in my view:
> For a character to be represented with a number, it is encoded as one or more bytes.  Each byte is a sequence of bits.  Each bit is 0 or 1.

Because the main point for this audience is that each character gets a number, like 65.

The fact that 65 in decimal can also be represented as 1000001 in binary or 41 in hex is irrelevant.

Because that's explaining how integers are encoded, which is out of scope.

In any case, we could add a column to the table to show the binary.

bittlingmayer:
Thanks for this knowledgeable contribution to a fundamental topic!

As Cecilia hinted, we have to keep the audience in mind:

People who:
- don't know what a string is:
- are accessing this page via machine translation

So we only need to cover what is relevant about strings *with regard to machine translation*, not with regard to all computer science and computer engineering.

Basically, non-technical people use words like "segment" or "sentence", but they may hear the word "string", which is more precise, but doesn't necessarily map to one thing in the product/workflows they know.

So how can we help them understand just enough and get on their way?

PR #355, opened 2022-11-28
Add APIs and languages counts to homepage

bittlingmayer:
Add the counts of APIs and languages to the homepage.

These are the most interesting articles actually.

We can make it dynamic later, like we have for contributors and contributions.

### DIFF
--- a/index.md
+++ b/index.md
@@ -13,6 +13,21 @@ seo:
 
 The content covers everything about machine translation, from products to research, and from history to news.
 
+<div style="display: flex; justify-content: center; gap: 20%;">
+  <div>
+    <center>
+      <h1>44</h1>
+      <a href="/apis">APIs</a>
+    </center>
+  </div>
+  <div>
+    <center>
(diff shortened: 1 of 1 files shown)


bittlingmayer:
Will look similar to this:

Screenshot 2022-11-28 at 10 13 15

but no emoji, and hyperlinked.

index.md
```diff
@@ -13,6 +13,21 @@ seo:
 
 The content covers everything about machine translation, from products to research, and from history to news.
 
+<div style="display: flex; justify-content: center; gap: 20%;">
+  <div>
+    <center>
+      <h1>44</h1>
+      <a href="/apis">APIs</a>
```
cefoo, line 20: Could we perhaps link them to `apis.md` and `languages.md`?

bittlingmayer: This effectively does.  Since it's HTML not markdown, it makes more sense to just use the final path, like we did further below.

PR #358, opened 2022-11-29
Add integrations to APIs

bittlingmayer:
TODO: Change 'engine' to 'api' in integrations.yml
TODO: Add to /_layouts/api.html

# Description

## Type of PR

- Edits _[article name]_
- Other _[please describe]_

### Checklist:

- [ ] I have read the [contributing guidelines](contributing.md).

### DIFF
--- a/_data/integrations.yml
+++ b/_data/integrations.yml
@@ -4,31 +4,31 @@
 #   url: <string> # link
 #   types: [ 'tms' | 'cat', ... ] # Functionality that it has
 #   open_source: true | false
-#   engine_integrations:
-#      - <string>: # slug of the corresponding engine id
-#          custom: true | false # Is the engine custom model available from the tool?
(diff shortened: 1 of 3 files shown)


cefoo:
I feel like we could improve the info distribution in API pages.
As is now, the `Integrations` list comes after the supported languages list, and there's a risk that readers looking for specific information might not scroll all the way down to the bottom of the page.

Perhaps we could:

- Move `Integrations` and place it before `Language support`.
- Create a TOC before `Language support` so that readers know that there is another section after the language list.

bittlingmayer:
Right, languages list can be very long.

For now, I put a link to integrations in the intro section, and in the infobox.  It's probably too much.  Maybe TOC is better.

My bet is that once it's live, we'll play around with it a bit.

PR #359, opened 2022-11-29
Add Languages link to APIs

bittlingmayer:
We already have the equivalent link in the other direction.

### DIFF
--- a/apis/apis.md
+++ b/apis/apis.md
@@ -7,7 +7,9 @@ description: Machine translation APIs
 
 There is a growing set of machine translation engines available.
 
-The most popular engines are available as self-serve APIs and support more than 100 languages and more than 10000 language pairs.  By 2022, there was a self-serve API that supported more than 300 languages.
+The most popular engines are available as self-serve APIs and support more than 100 [languages](/languages/languages.md) and more than 10000 language pairs.
+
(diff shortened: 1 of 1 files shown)


apis/apis.md
```diff
@@ -7,7 +7,9 @@ description: Machine translation APIs
 
 There is a growing set of machine translation engines available.
 
-The most popular engines are available as self-serve APIs and support more than 100 languages and more than 10000 language pairs.  By 2022, there was a self-serve API that supported more than 300 languages.
+The most popular engines are available as self-serve APIs and support more than 100 [languages](/languages/languages.md) and more than 10000 language pairs.
```
cefoo, line 10: Should the hover text explanation of what "self-serve" is be added here?

I'm not sure this is the way to go:

```
The most popular engines are available as <a title="Purchase process that doesn't require human assistance">self-serve</a> APIs
```

bittlingmayer: I googled it and the results are fairly unambiguous.  I'd avoid responsibility for defining things that are well-defined elsewhere (i.e. not specific to machine translation, or don't have a specific relevance and details in the context of machine translation, the way that e.g. byte-pair encoding or data confidentiality does).

PR #364, opened 2022-12-03
Language and API generation fixes

bittlingmayer:
There are some important fixes in the dataset here.  I was able to find a few errors, both from the APIs (e.g. `cz` instead of `cs` from Omniscient), and our own errors.

I added some validation logic to generate.py.  Overall, it needs to be simplified, but today is not the day for that.

I also added a few more languages.


_data/api-language.yml
```diff
@@ -16,6 +16,7 @@ microsoft:
   prs: fa-af # Macrolanguage, Microsoft for Dari
   fil: tl # Macrolanguage, Microsoft for Tagalog
   mww: hmn # Macrolanguage, Microsoft for Hmong
+  cz: cs # Czech, Omniscien
```
bittlingmayer, line 19: They fixed this, but it could happen again.

_data/apis.yml
```diff
@@ -208,7 +208,7 @@
   urls:
    - https://omniscien.com
    - https://omniscien.com/machine-translation/supported-languages/
-  languages: [ ar, be, bg, bn, ca, cz, da, de, el, en, es, et, fa, fi, fr, fr, ga, gu, he, hi, hr, hu, id, is, it, ja, km, ko, ku, lb, lt, lv, ms, mt, my, nl, 'no', pa, pl, pt, pt, ro, ru, si, sk, sl, so, sq, sr, sv, sw, ta, th, tk, tl, tr, uk, ur, vi, zh, zh ]
+  languages: [ ar, be, bg, bn, ca, cs, da, de, el, en, es, et, fa, fi, fr, fr-ca, ga, gu, he, hi, hr, hu, id, is, it, ja, km, ko, ku, lb, lt, lv, ms, mt, my, nl, 'no', pa, pl, pt-pt, pt-br, ro, ru, si, sk, sl, so, sq, sr, sv, sw, ta, th, tk, tl, tr, uk, ur, vi, zh-cn, zh-tw ]
```
bittlingmayer, line 211: Odd that they did not have locales.  We should probably add a check for repeated language codes in the list.

PR #369, opened 2022-12-20
Add stock domains to APIs

cefoo:
# Add stock domain-specific engines to APIs

## Type of PR

- Creates the article `stock-domains.yml`

### Checklist:

### Criteria

- For APIs where domains are insinuated or unspecified, or no concrete information about whether they offer domain-specific baseline models or not, I have set the `stock_api_domains` as `null`. For APIs that do not offer domain-specific models, I have set `stock_api_domains` as `false`.
- I have not added URLs where the info URL is the same as the one we have for the API itself, assuming we will display this domain information in the same page.

### DIFF
--- a/_data/stock-domains.yml
+++ b/_data/stock-domains.yml
@@ -1,0 +1,250 @@
+---
+
+- id: aisa
+  stock_api_domains:
+    - e-commerce
+    - gaming
+
+- id: alexa-ai
+  stock_api_domains:
+    - finance
+    - legal
+    - government
+  urls:
+    - https://alexatranslations.com/financial/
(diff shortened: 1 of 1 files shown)


bittlingmayer:
What do you think about calling these custom_domains?

cefoo:
> What do you think about calling these custom_domains?

I find it a little confusing as, from my understanding, the domains mentioned here are available as baseline models.

bittlingmayer:
And why not just put these in apis.yml?

cefoo:
OK, I'll add both `domains` and `privacy` in `apis.yml`, and continue this PR in #370

PR #370, opened 2022-12-22
Create privacy.yml

cefoo:
# Add data privacy information to APIs

Some providers' websites contain privacy information that just covers the use of the website, and not how they use the user's MT input data.
I have added the value `null` to the `privacy_urls` key in those cases and whenever the data privacy information is not included in the API documentation.

### Note

I have chosen the term `Privacy`, but different providers use other terms, such as `Data security`, `Data privacy`, and `Data confidentiality`.

---

## Type of PR

- Creates the article privacy.yml

### Checklist:


bittlingmayer:
Why not just put this in api.yml?

cefoo:
> Why not just put this in api.yml?

OK, ready.

I closed the other PR ( #369 ) and edited apis.yml with both privacy and domains info in this PR.

### Comments


- In the future, I should list the language pairs in which these stock API domains are available.
- I added both `stock_api_domains_urls` and `privacy_url`. Although this may not be the best choice since now most APIs have three variables referring to URLs, I considered this the best choice. Having them separated, we can place these URLs in a different spot from where we list the info URLs. I imagine something like this:

---
No self-service

Customisation:
  
  - Adaptive
  - Glossary
  
[kantanai.io](https://www.kantanai.io/)
[kantanmt.docs.apiary.io/#reference](https://kantanmt.docs.apiary.io/#reference)

Privacy:
  - [kantanai.io/privacy-statement-2/](https://www.kantanai.io/privacy-statement-2/)

Stock API domains:
  - automobile
  - e-commerce
  - gaming
  - government
  - IT
  - patents
  - tourism
 
[kantanai.io/showcase/](https://www.kantanai.io/showcase/)

---

bittlingmayer:
LGTM

I'm not too convinced we want to bother with the stock domain URLs, because it seems like they're not always there, not consistent, not the only thing on the page...

So they'll be hard to maintain - hard to find and add, will break a lot.

bittlingmayer:
Eventually we should also think about how to make a fixed list of domains.

Ideally we can make a page for each domain.

But each API may use subtly different names, like "film" vs "subtitles".

cefoo:
> Eventually we should also think about how to make a fixed list of domains.
> 
> Ideally we can make a page for each domain.
> 
> But each API may use subtly different names, like "film" vs "subtitles".

They do, some do have different names for the same domain.
Perhaps we can have a specific yaml for domains, like we have for mapping languages?

PR #350, opened 2022-11-23
Update word-embeddings.md

maxiek0071:
Article about word embeddings

## Type of PR

- Creates the article concepts/word_embeddings.md

### Checklist:

- [x] I have read the [contributing guidelines](contributing.md).
- [x] I have followed the [style guide](http://machinetranslate.org/style).

PR #371, opened 2022-12-24
Add tokenization

### LINKED ISSUE #74: Article: Tokenisation

ktrnka:
# Description

Fixes #74 

## Type of PR

- Edits _Tokenisation_

### Checklist:

# Challenges
- word vs token: I've most commonly heard people describe punctuation as being words but not tokens and that's how I wrote it. Though there are situations in which they're used interchangeably, for instance in notation for language models.
- Non-space-segmented languages: I'm not up-to-date on these languages. Last I heard, word segmentation is still used for them but I've seen a fair amount of NLP research that works at the character level as well.
- The need for tokenisation: It's entirely possible that there could be a breakthrough in character-based models in the future, and the whole need for tokenisation may evaporate. I tried to adhere to the guidelines about being evergreen but I didn't see a way to do that without being overly prescriptive about the future of research. Similarly, I'd love to cite a publication showing that splitting on spaces leads to worse machine translations than Treebank tokenizer and similar ones in English, but I'm not aware of any such work -- if anyone knows some let me know!

### DIFF
--- a/customisation/tokenisation.md
+++ b/customisation/tokenisation.md
@@ -12,8 +12,8 @@ Example:
 - input: `"Yes, but why?"`
 - output: `["Yes", ",", "but", "why", "?"]`
 
-Machine translation algorithms typically involve association between tokens in one language and tokens in another language.
-Tokenisation is needed because machine translation cannot operate effectively on the space of all possible strings.
+Machine translation algorithms learn the association between words in one language and words in another language.
+Tokenisation is needed because machine translation cannot operate effectively on the space of all possible strings, but can operate effectively on the space of all possible words.
 
 ## Languages
 The most basic tokenisation algorithm is splitting the string on whitespace.
--- a/customisation/tokenisation.md
+++ b/customisation/tokenisation.md
@@ -5,22 +5,24 @@ title: Tokenisation
 description: Splitting a string into a sequence of tokens
 ---
 
-Tokenisation is the process of splitting a string into a sequence of tokens.
-Tokens include words as well as punctuation characters.
+Tokenisation is the process of splitting a string into a sequence of substrings called tokens.
+A token is typically an atomic unit of meaning such as a word or punctuation character.
 
 Example:
 - input: `"Yes, but why?"`
 - output: `["Yes", ",", "but", "why", "?"]`
 
-Machine translation algorithms learn the association between words in one language and words in another language.
-Tokenisation is needed because machine translation cannot operate effectively on the space of all possible strings, but can operate effectively on the space of all possible tokens.
+Tokenization has a large effect on the vocabulary size of the machine translation system.
+For example, if punctuation is not split off from words the vocabulary size will be much larger.
+If the vocabulary size is larger, much more data is necessary to learn a machine translation model that will generalize.
 
-## Languages
+## Challenges
 The most basic tokenisation algorithm is splitting the string on whitespace.
-Splitting on spaces is a reasonable baseline for many languages.
-However, there are several languages that do not separate words with whitespace, such as Chinese, Japanese, Korean, and Thai.
-Machine translation of non-space-segmented languages often relies on more complex word segmentation algorithms.
+Splitting on spaces is a reasonable baseline for many languages but will lead to an overly large vocabulary in other languages.
+Two types of challenging languages are non-space-segmented languages and agglutinative languages.
 
-## See also
+Non-space-segmented languages include Chinese, Japanese, Korean, and Thai.
+Tokenization is typically addressed by use of more advanced algorithms in these languages, often called word segmentation algorithms.
 
-- [Byte-pair encoding](/approaches/byte-pair-encoding.md): Byte-pair encoding is related because it also splits strings into smaller units.
+Agglutinative languages include Finnish and Hungarian.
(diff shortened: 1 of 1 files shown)


ktrnka:
CC @cefoo

customisation/tokenisation.md
```diff
@@ -1,7 +1,26 @@
...
 grand_parent: Building and research
 parent: Customisation
-layout: coming_soon
 title: Tokenisation
-description:
+description: Splitting a string into a sequence of tokens
 ---
+
+Tokenisation is the process of splitting a string into a sequence of tokens.
+Tokens include words as well as punctuation characters.
```
cefoo, line 9: Do you think adding a sentence explaining what a `token` is would be useful here?
Something like...

```
Tokens are string elements that carry meaning.
Tokens include words as well as punctuation characters.
```

cefoo: Perhaps we could differentiate the meaning of `token` vs `word type`, so that readers know that `tokens` include repetitions.

In the [PR for word embeddings](https://github.com/machinetranslate/machinetranslate.org/pull/350), we mention both `vector` and `word embedding` concepts. Do you think we can link `word embedding` here, if we mention `word type`? Or perhaps that deserves a new article altogether.

customisation/tokenisation.md
```diff
@@ -1,7 +1,26 @@
...
+- input: `"Yes, but why?"`
+- output: `["Yes", ",", "but", "why", "?"]`
+
+Machine translation algorithms learn the association between words in one language and words in another language.
+Tokenisation is needed because machine translation cannot operate effectively on the space of all possible strings, but can operate effectively on the space of all possible tokens.
+
+## Languages
+The most basic tokenisation algorithm is splitting the string on whitespace.
+Splitting on spaces is a reasonable baseline for many languages.
+However, there are several languages that do not separate words with whitespace, such as Chinese, Japanese, Korean, and Thai.
```
cefoo, line 21: I would rename this section as `## Challenges` and also mention agglutinative languages.

cefoo (commented):
Hi @ktrnka!

Thank you so much for your contribution!

I added some minor suggestions, let me know what you think!

cefoo:
> word vs token: I've most commonly heard people describe punctuation as being words but not tokens and that's how I wrote it. Though there are situations in which they're used interchangeably, for instance in notation for language models.

I think the way you wrote it is fine.

> Non-space-segmented languages: I'm not up-to-date on these languages. Last I heard, word segmentation is still used for them but I've seen a fair amount of NLP research that works at the character level as well.

Just mentioning that these languages are challenging and why is enough for now, IMHO.

> The need for tokenisation: It's entirely possible that there could be a breakthrough in character-based models in the future, and the whole need for tokenisation may evaporate. I tried to adhere to the guidelines about being evergreen but I didn't see a way to do that without being overly prescriptive about the future of research. Similarly, I'd love to cite a publication showing that splitting on spaces leads to worse machine translations than Treebank tokenizer and similar ones in English, but I'm not aware of any such work -- if anyone knows some let me know!

This article just describes what `tokenisation` is. I believe that, in case of a breakthrough, we can update the article or add the relevant information to the languages articles.
If you check [here](https://github.com/machinetranslate/machinetranslate.org/blob/master/_data/languages.yml), you'll notice we have information about languages typology, which we still haven't published.
Perhaps it would be more adequate to create a typology article, listing languages that are characterized by it and methods to tackle it in machine translation. What do you think?

customisation/tokenisation.md
ktrnka: Oh, thanks I see it's problematic to introduce the concept of a token that way, let me try revising that.

I'll see if I can downplay the distinction between `token` and `word` or `word instance` because most people use the terms interchangeably

customisation/tokenisation.md
ktrnka: Yeah that'd make sense, and would make it easier for me to introduce BPE. I'll try rephrasing the choice of tokenization algorithm to affect the vocabulary size, and therefore the curse of dimensionality, and therefore MT quality. I hope if I do that right, it'll be a natural transition into agglutinative, non-space-segmented languages, and others

ktrnka:
Ok, I updated - could you take another look?

Re: Typology article, yes, once we have one it'd be nice to link to a list of agglutinative languages. I didn't see anything in the file about space segmentation though so it probably wouldn't help there. I'd love to also mention German and languages like it, because those are not agglutinative but those compounds are problematic for modeling.

I didn't see any need to link to word embeddings.

customisation/tokenisation.md
cefoo: Great, thanks!

cefoo:
> Re: Typology article, yes, once we have one it'd be nice to link to a list of agglutinative languages. I didn't see anything in the file about space segmentation though so it probably wouldn't help there. I'd love to also mention German and languages like it, because those are not agglutinative but those compounds are problematic for modeling.

Great, thanks!
I think that, as the site continues to grow, we can always go back and hyperlink terms/concepts, if necessary.

Yeah, adding German would make it even more clear!

ktrnka (commented):
Yep @cefoo I took a look and don't see any issues

PR #375, opened 2022-12-28
Update api-language.yml

bittlingmayer:


cefoo:
What's the logic behind the key-value location? Are you grouping the codes that have comments at the beginning, and then it is alphabetic?

bittlingmayer:
Yes, exactly, any funny business should be at the top.  The rest is very boring.

_data/api-language.yml
```diff
@@ -17,6 +16,9 @@ microsoft:
   fil: tl # Macrolanguage, Microsoft for Tagalog
   mww: hmn # Macrolanguage, Microsoft for Hmong
   cz: cs # Czech, Omniscien
+  me: cnr # Montenegrin, Niutrans
+  mo: mn # Mongolian, Niutrans
+  jy: ka # Georgian, Niutrans
```
bittlingmayer, line 21: Arguably, these should go under `niutrans:`

It probably doesn't hurt to have them here, but it's hard to imagine another provider doing this exact same terrible deviation from the standard.

cefoo: I'll merge. We can reorganize it later if we decide to go with that option.

PR #376, opened 2022-12-30
Create token.md

### LINKED ISSUE #111: Article: Token

ktrnka:
# Description

Fixes #111 by creating token.md. Also links from tokenisation.md into it. The content is minimal; I don't think there's really much to say about the concept of a token separate from variations in tokenisation algorithms.

## Type of PR

- Creates the article _Token_

### Checklist:

### DIFF
--- a/concepts/token.md
+++ b/concepts/token.md
@@ -8,4 +8,4 @@ description: A short sequence of characters
 A token is a short sequence of characters with meaning, most typically a word or punctuation character.
 Strings are split into tokens by [tokenisation](/customisation/tokenisation) algorithms.
 
-Examples: "this", "sequence", "?"
+Examples: “this”, “sequence”, “?”
--- a/concepts/token.md
+++ b/concepts/token.md
@@ -1,0 +1,11 @@
+---
+grand_parent: Resources
+parent: Concepts
+title: Token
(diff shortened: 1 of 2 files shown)


cefoo:
Thanks for the PR, @ktrnka!
I just changed from straight quotes to curly quotes.

concepts/token.md
```diff
@@ -0,0 +1,11 @@
...
+grand_parent: Resources
+parent: Concepts
+title: Token
+description: A short sequence of characters
+---
+
+A token is a short sequence of characters with meaning, most typically a word or punctuation character.
+Strings are split into tokens by [tokenisation](/customisation/tokenisation) algorithms.
+
+Examples: “this”, “sequence”, “?”
```
bittlingmayer, line 11: @cefoo I think we should add something to the style guide that stuff like this should delimited `with backticks` so that later we can avoid translating it.

> Examples
> > `this`, `sequence`, `?`

And maybe in blockquote?

cefoo: @bittlingmayer

I would choose backticks.

We widely use blockquote in Events, Calls for papers, and in our Meetup 2 page, specifically for quotes, so I would leave that format for those cases.

I'll update the Style Guide.

PR #378, opened 2022-12-31
Create vocabulary.md

### LINKED ISSUE #372: Article: Vocabulary
**Article title:** Vocabulary

**Directory:** Concepts

**Description**: Training data vocabulary for machine translation

**Definition**: **Vocabulary** is a list of all the unique words (word types) that occur in the training data.

**Linked concepts**: Tokenisation, neural machine translation, word embedding

ktrnka:
# Description

Fixes #372 - adding an article for vocabulary.

## Type of PR

- Creates the article _Vocabulary_

### Checklist:

## Challenges (could use advice on)

- If I remember right, neural models slow down with vocabulary due to the size of the softmax in the output, though I'm not sure if that's still a problem these days. I vaguely remember that the ways to share embedding weights between the decoder input & output may eliminate that problem, at least in training
- I only recently learned that translation of single-occurrence words can be worse than OOVs from <div class="csl-entry">Koehn, P., &#38; Knowles, R. (2017). <i>Six Challenges for Neural Machine Translation</i>. 28–39. https://doi.org/10.48550/arxiv.1706.03872</div>
- It might be nice to add typical vocabulary sizes for some languages as examples, though that may not remain evergreen (it's already significantly affected by BPE)
- I referenced the concept of an open-class word type which may confuse readers. Not sure how else to phrase that, or if it's fine and better to define it later?

## Issues to add if merged

- Add an article on copy/pointer mechanism
- Clarify existing articles on typical preprocessing of numbers -- ASR systems tend to spell them out like nineteen fifty instead of 1950 when training then reverse it when recognizing, maybe MT systems do the same?


concepts/vocabulary.md
```diff
@@ -0,0 +1,31 @@
+---
+grand_parent: Resources
+parent: Concepts
+title: Vocabulary
+description: The set of supported words
+---
+
+The **vocabulary** of a model is the set of supported words.
+In a machine translation system, the vocabulary may refer to the set of words supported in either the input or the output.
+A word that's supported in the output is a word that can be generated by the translation model.
```
cefoo, line 10: According to our [Style guide](https://machinetranslate.org/style#apostrophes-and-quotation-marks), we should use the curly apostrophe. However, I think we can eliminate the `that's` altogether.

concepts/vocabulary.md
```diff
@@ -0,0 +1,31 @@
...
+A word that's supported in the input is a word that can be translated.
```
cefoo, line 11: Shall we change the order and first define `input-supported word` and second `output-supported word`? I know it's a tiny edit, but it would follow the order introduced by the previous sentence.

concepts/vocabulary.md
```diff
@@ -0,0 +1,31 @@
...
+- A large vocabulary will allow the system to learn to translate more words, but makes the model much larger and slower.
+- In [neural machine translation](/approaches/neural-machine-translation.md), it can be challenging to share the vocabulary between the decoder inputs and decoder outputs.
+- If pretrained word embeddings are used, the vocabulary is already determined and not trivial to update with new data.
+- Open-class words present challenges, such as proper names and numbers.
+- Agglutinative and highly inflected languages have very large vocabularies, and may require massive amounts of training data to observe enough of the vocabulary.
+
+## Dependencies
+
+- The vocabulary is affected by choice of [tokenisation algorithm](/customisation/tokenisation.md) or the use of subword models such as [byte-pair encoding](/approaches/byte-pair-encoding.md). In the case of byte-pair encoding, this can cause vocabulary size to become a hyperparameter that affects the generalization of the model.
+- The vocabulary is affected by normalization of capitalization and accents.
```
cefoo, line 31: To keep consistency with the rest of the website, we should use [UK spellings](https://machinetranslate.org/style#default-to-wikipedia-style).

- Generalisation
- Normalisation
- Capitalisation

concepts/vocabulary.md
```diff
@@ -0,0 +1,31 @@
...
+Words that are not in the vocabulary are called out-of-vocabulary (OOV).
+
+The vocabulary of a language is the set of all possible words in the language.
+
+The vocabulary of a pretrained [word embedding model](/concepts/word-embeddings.md) is the set of all words with defined embeddings.
+
+## Challenges
+
+- A large vocabulary will allow the system to learn to translate more words, but makes the model much larger and slower.
+- In [neural machine translation](/approaches/neural-machine-translation.md), it can be challenging to share the vocabulary between the decoder inputs and decoder outputs.
```
cefoo, line 23: I would eliminate this sentence as it is hard to understand by non-developers/researchers.
Alternatively, we could further explain what sharing vocabulary between the inputs and outputs means and why it is done, but I feel it would make this article too granular.

cefoo:
Thanks for the PR, @ktrnka! 🎉

> ## Challenges (could use advice on)
> 
> If I remember right, neural models slow down with vocabulary due to the size of the softmax in the output, though I'm not sure if that's still a problem these days. I vaguely remember that the ways to share embedding weights between the decoder input & output may eliminate that problem, at least in training

I am not the person who can help with this question, but perhaps @bittlingmayer or @zouharvi can?
From what I've read, this is all correct.

> I only recently learned that translation of single-occurrence words can be worse than OOVs from Koehn, P., & Knowles, R. (2017). Six Challenges for Neural Machine Translation. 28–39. https://doi.org/10.48550/arxiv.1706.03872

I think @bittlingmayer or @zouharvi can help here too.

> It might be nice to add typical vocabulary sizes for some languages as examples, though that may not remain evergreen (it's already significantly affected by BPE)

To introduce non-evergreen information, we can present the information as true at a certain point in time.
So, for instance, we could say:

```
By 2022, typical vocabularies were...
```

We have resorted to this type of wording in some other parts of the website.

> I referenced the concept of an open-class word type which may confuse readers. Not sure how else to phrase that, or if it's fine and better to define it later?

I think it's fine as is. Perhaps we can say "Open class or content words"? But I'm not sure if they are equivalent.


> 
> ## Issues to add if merged
> 
> Add an article on copy/pointer mechanism

Should we add an issue about it?

> Clarify existing articles on typical preprocessing of numbers -- ASR systems tend to spell them out like nineteen fifty instead of 1950 when training then reverse it when recognizing, maybe MT systems do the same?

I think this is worth mentioning in the speech translation article. What do you think?

concepts/vocabulary.md
ktrnka: Easier to eliminate it yeah.

I'm surprised that the style guide is meant to apply to word-internal quotes though, especially since the site is meant to be sent through MT. Many NLP systems don't handle "smart quotes" very well because it makes it a different token altogether unless it's merged in preprocessing.

concepts/vocabulary.md
ktrnka: Yeah it would really expand the scope too much. That sort of thing belongs in the NMT article if anywhere

concepts/vocabulary.md
cefoo: You are absolutely right, I must fix that.

ktrnka:
Ok I think I addressed most of the issues. 

I don't think we need to do anything about the NMT references because we removed the only NMT reference.

The one remaining thing is to indicate some common vocabulary sizes -- I'll pull some WMT 2022 papers and see if I can get ranges there. One caveat is that I think everyone uses BPE these days and those vocab sizes would be misleading if anyone tries to use them for non-BPE vocabs.

ktrnka:
Ok ready for another review!

zouharvi:
Hi, what a coincidence that I'm currently working on the problem of vocabulary size. :-)

Very common for small to mid models is something between 8k and 16k. Usually the input and output vocabularies are joint (i.e. the subworder/tokenizer is trained on both of them together).

[citation](https://arxiv.org/abs/2004.02334)

The vocabulary size is a tradeoff. If you increase it, you'll get shorter sentences but larger output vectors over which you'll have to softmax. With higher vocabularies you'll also get subwords which are not very frequent.

cefoo:
> Hi, what a coincidence that I'm currently working on the problem of vocabulary size. :-)
> 
> Very common for small to mid models is something between 16k and 32k. Usually the input and output vocabularies are joint (i.e. the subworder/tokenizer is trained on both of them together).
> 
> [citation](https://arxiv.org/abs/2004.02334)
> 
> The vocabulary size is a tradeoff. If you increase it, you'll get shorter sentences but larger output vectors over which you'll have to softmax. With higher vocabularies you'll also get subwords which are not very frequent and that's problematic for training (sparsity).

Thanks for your input, @zouharvi! :)

PR #383, opened 2023-01-04
Update Burmese with a note on Zawgyi

### LINKED ISSUE #380: Update burmese with information about Zawgyi
It might be good to explain Zawgyi for Burmese on the appropriate language page. It's supported by XL8. More info: https://en.wikipedia.org/wiki/Zawgyi_font

From another issue:
> It's tempting to bring up Zawgyi for Burmese which isn't a script variant but a character encoding variant (it's a slight modification of Unicode). I could add that if it's valuable -- I only see it in api.yaml for one system and I'd guess the other systems _probably_ use Unicode
> 

Perhaps we can include a short note in the relevant language page?
Although it's not clear right now, we want to include specific information about each language in the languages pages.
If you check [Russian](https://machinetranslate.org/russian) or [Turkish](https://machinetranslate.org/turkish), this will be clearer.
Do you think it would be more relevant to insert this info in the language-specific pages?

_Originally posted by @cefoo in https://github.com/machinetranslate/machinetranslate.org/issues/377#issuecomment-1369915146_

ktrnka:
# Description

Fixes #380 by adding a section describing the Unicode vs Zawgyi difference and how it could affect training of MT.

## Type of PR

- Edits _burmese.md_

### Checklist:

### DIFF
--- a/languages/burmese.md
+++ b/languages/burmese.md
@@ -54,5 +54,5 @@ Burmese characters can be represented in Unicode or Zawgyi.
 Zawgyi is a variation on Unicode which interprets some code points differently but otherwise encodes code points with standard encodings such as UTF-8.
 If Zawgyi text is interpreted as Unicode, some characters will appear garbled.
 
-When building training data for machine translation from web text, the data will include a mixture of Burmese Unicode and Burmese Zawgyi.
+Burmese text data from across the web is a mixture of Burmese Unicode and Burmese Zawgyi encoding.
 This can lead to fragmentation of some of the vocabulary unless the data is first normalized.
--- a/languages/burmese.md
+++ b/languages/burmese.md
@@ -48,3 +48,11 @@ supported_apis:
 
 ---
 
+## Character encoding
+
+Burmese characters can be represented in Unicode or Zawgyi.
+Zawgyi is a variation on Unicode which interprets some code points differently but otherwise encodes code points in the same way with UTF-8 and other encodings.
+If Zawgyi text is interpreted as Unicode, some characters will appear garbled.
+
+When building training data for machine translation from web text, the data will include a mixture of Burmese Unicode and Burmese Zawgyi.
+This can lead to fragmentation of some of the vocabulary unless the data is first normalized.
--- a/languages/burmese.md
+++ b/languages/burmese.md
@@ -51,7 +51,7 @@ supported_apis:
 ## Character encoding
 
 Burmese characters can be represented in Unicode or Zawgyi.
-Zawgyi is a variation on Unicode which interprets some code points differently but otherwise encodes code points in the same way with UTF-8 and other encodings.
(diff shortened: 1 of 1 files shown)


languages/burmese.md
```diff
@@ -48,3 +48,11 @@ supported_apis:
 
 ---
 
+## Character encoding
+
+Burmese characters can be represented in Unicode or Zawgyi.
+Zawgyi is a variation on Unicode which interprets some code points differently but otherwise encodes code points in the same way with UTF-8 and other encodings.
```
bittlingmayer, line 54: Maybe *as* not *with*?

> the same way as UTF-8.

languages/burmese.md
```diff
@@ -48,3 +48,11 @@ supported_apis:
...
+If Zawgyi text is interpreted as Unicode, some characters will appear garbled.
+
+When building training data for machine translation from web text, the data will include a mixture of Burmese Unicode and Burmese Zawgyi.
```
bittlingmayer, line 57: Maybe just

> Burmese text data from across the web is a mixture of Burmese Unicode and Burmese Zawgyi encoding.

languages/burmese.md
```diff
@@ -48,3 +48,11 @@ supported_apis:
...
+This can lead to fragmentation of some of the vocabulary unless the data is first normalized.
```
bittlingmayer, line 58: Is mixed encoding perhaps desirable when going *from* Burmese to e.g. English, but not from e.g. English *to* Burmese?

bittlingmayer (approved):
Awesome contrib!

Useful gotcha and conforms to our style.

I have only a few rewording suggestions and one question.

Later we should think about what other languages have this type issue, and if it's worth putting this info in the structured data for all of them.

languages/burmese.md
ktrnka: Looking at it more, I'll simplify further and just say "with UTF8 and others". I was trying to avoid saying that it used UTF-8 because I wasn't sure if UTF-8 implies Unicode code points. Wikipedia doesn't seem to tie them together too carefully though so I think it's better to simplify

languages/burmese.md
ktrnka: I hadn't considered that... I _think_ it'd be desirable yes, because end users may not know the encoding of their text so it'd be good to support both.

ktrnka:
I made the changes so I'll merge!

> Later we should think about what other languages have this type issue, and if it's worth putting this info in the structured data for all of them.

Burmese is the only language I know about with this issue, out of the 80 or so languages and variants we supported in the Nuance mobile keyboard SDK. It wouldn't surprise me if there are others, I just don't know about them.

The closest thing I can think of happens in Romanian, where there's fragmentation because some people type s-with-sedilla and others type s-with-comma to mean the same thing. If I remember the story correctly, there was some mixup with Adobe fonts or the Unicode spec lagged behind computer adoption which led to it.

PR #388, opened 2023-01-09
Add sentence splitting

### LINKED ISSUE #174: Article: Sentence splitting

ktrnka:
# Description

Fixes #174 by adding an article on sentence splitting

## Challenges - could use some perspectives on this

- A beginner would probably want to know why we split into sentences. I tried to briefly explain it but didn't do the best job (I just said it's for efficiency, which is true but glosses over details). More than anything sentence splitting feels like a good compromise between making it easy/fast for MT to learn and not needing to do tons of annotation work. Alignment at the word level might enable better MT but would probably take ages to annotate, and would likely be very noisy. Alignment at the document level is less work but MT needs to consider so many more possible word-level alignments (though it's fairly common, just with a sentence alignment stage first).
- I didn't want to get too much into depth about challenges in segmentation or various algorithms

## Type of PR

- Creates the article _Sentence splitting_

### Checklist:

### DIFF
--- a/concepts/sentence-splitting.md
+++ b/concepts/sentence-splitting.md
@@ -4,3 +4,8 @@ parent: Concepts
 title: Sentence splitting
 description: Splitting a text into a sequence of sentences
 ---
+
+**Sentence splitting** is the process of converting a text into a sequence of sentences.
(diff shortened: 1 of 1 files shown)


ktrnka:
Nah, feel free to merge. If we come up with better wording later we can revise

PR #390, opened 2023-01-11
Create language identification

### LINKED ISSUE #95: Language identification

ktrnka:
# Description

Fixes #95 by creating the language identification page.

## Type of PR

- Creates the article _Language identification_

### Checklist:

### DIFF
--- a/customisation/language-identification.md
+++ b/customisation/language-identification.md
@@ -12,5 +12,5 @@ Language identification tools are also used to automatically identify the source
 ## Challenges
 
 - Similar languages may be mistaken for one another, for example [Dutch](/languages/dutch.md) and [Afrikaans](/languages/afrikaans.md).
-- Language identification is more challenging on short pieces of text.
+- Language identification is harder on shorter texts because there may not be enough language-specific information. Longer texts have more chances of having language-specific characters and words making it easier to identify.
 - If a particular [locale](/applications/advanced-concepts/locale.md) is not supported in the language identification tool, it may be misclassified as another, similar locale. For example, if a language identification tool does not support [Serbian](/languages/serbian.md) with the Cyrillic script, it may be misclassified as [Macedonian](/languages/macedonian.md).
--- a/customisation/language-identification.md
+++ b/customisation/language-identification.md
@@ -11,6 +11,6 @@ Language identification tools are also used to automatically identify the source
 
 ## Challenges
 
-- Similar languages may be mistaken for one another, for example Dutch and Afrikaans.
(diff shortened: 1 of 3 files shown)


customisation/language-identification.md
```diff
@@ -0,0 +1,16 @@
...
+---
+
+**Language identification** tools identify the language of an unknown text.
+In machine translation, they are used in building monolingual [training data](/customisation/training-data.md) from large collections of unidentified texts such as webpages.
+They are also used to automatically identify the source language in many machine translation systems when the source language is unspecified.
+
+## Challenges
+
+- Similar languages may be mistaken for one another, for example Dutch and Afrikaans.
+- Language identification is more challenging on short pieces of text.
```
cefoo, line 15: Perhaps we can extend this sentence explaining why.
 "Language identification is more challenging on short pieces of text because many languages can include the same word, and..."

customisation/language-identification.md
```diff
@@ -0,0 +1,16 @@
...
+- If a particular [locale](/applications/advanced-concepts/locale.md) is not supported in the language identification tool, it may be misclassified as another, similar locale. For example, if a language identification tool does not support Serbian with the Cyrillic script, it may be misclassified as Macedonian.
```
cefoo, line 16: I think that, since we are mentioning locales here, perhaps at the language identification definition above, we should mention it as well.

customisation/language-identification.md
```diff
@@ -0,0 +1,16 @@
+---
+grand_parent: Building and research
+parent: Customisation
+title: Language identification
+description: Identifying the language of a text
+---
+
+**Language identification** tools identify the language of an unknown text.
+In machine translation, they are used in building monolingual [training data](/customisation/training-data.md) from large collections of unidentified texts such as webpages.
+They are also used to automatically identify the source language in many machine translation systems when the source language is unspecified.
```
cefoo, line 10: Shall we mention something like...?

"Some language identification methods are n-gram probability, dictionaries, and more."

I fear this may end up not being evergreen (or is even not evergreen now), but we can fix it by saying "By 2022, ...".

cefoo (commented):
Hi @ktrnka!

Thanks for the PR! ✨

customisation/language-identification.md
ktrnka: Yeah I couldn't decide either way, but if you think it'd be helpful I'm happy to add that.

cefoo: I think it can be interesting to mention it as long as it adds info that's accessible to all our readers profile.
Perhaps `n-gram probability` can be tricky, but if we link to the `n-gram` article, we create a nice reading path for the audience. 
What do you think?

PR #391, opened 2023-01-16
Create: Interactive machine translation

### LINKED ISSUE #333: Article: Predictive MT
**Article title:** *Predictive MT*

**Directory:** *Customisation*

**Definition**: Recalculation of the MT as the translator makes changes

ktrnka:
# Description

Fixes #333 by making a new page.

## Challenges

- If I understand publications right, IMT was less efficient for professional translators with SMT and now it can be more efficient with NMT. The findings depend on the user interface though. It didn't feel right to include that, but it's useful history to know.
- I included the one big challenge from <div class="csl-entry">Foster, G., Isabelle, P., &#38; Plamondon, P. (1997). Target-Text Mediated Interactive Machine Translation. <i>Machine Translation</i>, <i>12</i>(1–2), 175–194. https://doi.org/10.1023/a:1007999327580</div> but I'm not sure if it's helpful or just confusing

## Type of PR

- Creates the article _Interactive machine translation_

### Checklist:


bittlingmayer:
Microsoft also calls it "interactive": https://slate.com/technology/2023/01/storyweaver-ai-translation-tools-language-preservation.html

ktrnka:
Hmm, they call it interactive neural machine translation prediction in there, and abbreviate it INMT. I've seen interactive translation prediction (ITP) also in the user study and the Casmacat work, though Koehn was in both works so it's more like one "vote" for that name.

I think "interactive machine translation prediction" or "interactive translation prediction" are probably the best names in terms of popularity and understandability. What do you think?

bittlingmayer:
Thanks for laying out the options

I'd vote for "interactive translation prediction".

Reasoning:
- In my view, "prediction" already implies machine i.e. automatic.  Adding "machine" would be a bit like calling monolingual keyboard suggestions "machine auto-suggestion" or "automatic autocorrect"...
- It's relatively established.  "interactive machine translation prediction" has only 1 occurrence on the internet.
- It's shorter.

bittlingmayer:
I don't have a strong feeling about the primary name, I just want to mention how we deal with multiple names:
 
- All variants are at the beginning, in bold.
- We can use the abbreviation once in parentheses, also in bold.
- We do not use the abbreviation elsewhere in the article.

> **Interactive machine translation** (**IMT**) or **interactive translation prediction** (**ITP**) helps human translators type translations with a predictive text input system.
> Unlike traditional predictive text, interactive machine translation uses the machine translation model to provide better completions of the sentence.

Obviously if we go with **interactive translation prediction** then it'll be slightly different.

ktrnka:
Sounds good -- I revised it to just be ITP

cefoo:
Thanks for the PR, @ktrnka!
I'll merge now.

PR #392, opened 2023-01-17
Update style guidance on acronyms and abbreviations

bittlingmayer:
# Description

More clarity on how we deal with acronyms

## Type of PR

- Style guide update

### Checklist:


zouharvi:
I don't think there's anything controversial, so precision-wise it's 100%. Whether we're omitting something I am not sure but it seems like a straightforward part of the guidelines.

PR #405, opened 2023-01-18
Update apis.md

bittlingmayer:
# Description

Better wording, mention TMS

## Type of PR

- Edits *APIs*

### Checklist:

### DIFF
--- a/apis/apis.md
+++ b/apis/apis.md
@@ -10,7 +10,7 @@ There is a growing set of machine translation APIs available.
 The most popular APIs are available as self-serve APIs and support more than 100 [languages](/languages/languages.md) and more than 10000 language pairs.
 By 2022, there was a self-serve API that supported more than 300 languages.
 
-Besides language and [locale](/building-and-research/locale.md) support, the APIs also vary by features like customisation, privacy policies and whether or not they are self-service products.
+Besides language and [locale](/building-and-research/locale.md) support, the APIs also vary by features like [customisation](/building-and-research/customisation.md), [privacy policies](/industry/data-confidentiality.md) and whether or not they are self-service products.
 Many of the APIs have been integrated into translation management systems (TMS) where human translators work to [post-edit](https://machinetranslate.org/post-editing) machine translation output.
 
 Most of the underlying machine translation systems are now based on [neural machine translation](/approaches/neural-machine-translation.md).
--- a/apis/apis.md
+++ b/apis/apis.md
@@ -5,12 +5,14 @@ title: APIs
 description: Machine translation APIs
 ---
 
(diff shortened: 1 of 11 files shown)


apis/apis.md
```diff
@@ -5,12 +5,14 @@ title: APIs
 description: Machine translation APIs
 ---
 
-There is a growing set of machine translation engines available.
+There is a growing set of machine translation APIs available.
 
-The most popular engines are available as self-serve APIs and support more than 100 [languages](/languages/languages.md) and more than 10000 language pairs.
+The most popular APIs are available as self-serve APIs and support more than 100 [languages](/languages/languages.md) and more than 10000 language pairs.  By 2022, there was a self-serve API that supported more than 300 languages.
```
cefoo, line 10: In our [style guide](https://machinetranslate.org/style#spacing), we established that sentences should be separated by a new line. I know the result is the same, but we should follow the style guide to unify.

cefoo: Should we link to Niutrans when mentioning this fact?
Although this can change in the future, by 2022 it was true.

I'm afraid that, if another API supports more languages in the future, this sentence would end up being misleading as the audience may think the current API that supports many languages is the one we are talking about.

apis/apis.md
```diff
@@ -5,12 +5,14 @@ title: APIs
...
 
-By 2022, there was a self-serve API that supported more than 300 languages.
+Besides language and locale support, the APIs also vary by features like customisation, privacy policies and whether or not they are self-service products.  Many of the APIs have been integrated into translation management systems (TMS) where human translators work to [post-edit](https://machinetranslate.org/post-editing) machine translation output.
```
cefoo, line 12: Let's link to the locale article, and perhaps privacy policies ([this article](/industry/data-confidentiality.md)) and products?

bittlingmayer:
I'm not too worried about the 300 languages not being evergreen, because it is a rare and notable event, and we can even make it dynamic before that happens.

I believe it will be something like this:
```
{{ site.data.apis | map: "languages" | map: "size" | sort | last }}
```
except that for some APIs we use language pairs.

apis/apis.md
bittlingmayer: I updated the style guide.

bittlingmayer: I'm not too worried about the 300 languages not being evergreen, because it is a rare and notable event, and we can even make it dynamic before that happens.

I believe it will be something like this:
```
{{ site.data.apis | map: "languages" | map: "size" | sort | last }}
```
except that for some APIs we use language pairs.

apis/apis.md
bittlingmayer: That's a different type of product

Issue #73, opened 2021-11-30
Transformers


cefoo:

lena-voita:
Is anyone working on this? I could take this one

bittlingmayer:
Some things to make sure of:
- Link *Transformers* and *Attention* to each other
- Make sure they're in the same directory (right now they're now
- Add *Attention is All you Need* to https://machinetranslate.org/resources/publications/#attention-mechanism

Other practical bits of knowledge we could add to both *Attention* and *Transformers*:
- A graphic, from Wikipedia or Google Research
- How adopted it is - that most MT systems use it, and that it's used outside of MT
- Who invented it and/or which institution it came out of
- Years of invention and adoption
- Good resources
   - https://lena-voita.github.io/nlp_course/seq2seq_and_attention.html
   - https://huggingface.co/docs/transformers/index

Meta tasks, out of the scope of this issue:
- Move publications out to the respective articles.

PR #414, opened 2023-01-25
Update transformers.md

liashahnazaryan:
# Description

Created content for the `Transformers` page

## Type of PR

- Creates the article _[Transformers]_

### Checklist:

### DIFF
--- a/approaches/transformers.md
+++ b/approaches/transformers.md
@@ -8,12 +8,15 @@ description: A deep learning model architecture for neural machine translation
 **Transformers** are a type of neural network architecture that relies only on the [attention](/concepts/attention.md) mechanism.
 It allows [neural machine translation](/approaches/neural-machine-translation.md) models to better handle the dependencies between the input and output sequences.
 
-The transformer was first introduced in a 2017 paper by Google researchers titled [*”Attention is all you need.”*](https://arxiv.org/pdf/1706.03762.pdf)
+The transformer was first introduced in the paper titled [*”Attention is all you need”*](https://arxiv.org/pdf/1706.03762.pdf) by Google Research in 2017.
 
 ## Architecture
 
 The transformer consists of an encoder and a decoder, which have several layers.
 These layers include sublayers of *self-attention* and *feed-forward neural networks*.
+
+> ##### Image from a Wikipedia [article](https://en.wikipedia.org/wiki/Transformer_(machine_learning_model))
+> <img title='Transformer model architecture' src='approaches/transformer-model-architecture.png' width='700' style='padding: 1em;' />
 
 ### Encoder
 
--- a/approaches/transformers.md
+++ b/approaches/transformers.md
@@ -8,12 +8,12 @@ description: A deep learning model architecture for neural machine translation
 **Transformers** are a type of neural network architecture that relies only on the [attention](/concepts/attention.md) mechanism.
 It allows [neural machine translation](/approaches/neural-machine-translation.md) models to better handle the dependencies between the input and output sequences.
 
-The Transformer was first introduced in a 2017 paper by Google researchers titled _"Attention Is All You Need."_
+The transformer was first introduced in a 2017 paper by Google researchers titled [*”Attention is all you need.”*](https://arxiv.org/pdf/1706.03762.pdf)
 
 ## Architecture
 
-The Transformer consists of an encoder and a decoder.
-Each consists of several layers, which include sublayers of *self-attention* and *feed-forward neural networks*.
+The transformer consists of an encoder and a decoder, which have several layers.
+These layers include sublayers of *self-attention* and *feed-forward neural networks*.
 
 ### Encoder
 
@@ -30,12 +30,11 @@ The decoding process is similar to the one used in the encoder, as the decoder u
 But in addition to the self-attention and feed-forward neural network layers, the decoder uses an additional mechanism called *masked self-attention*.
 It partially masks the output sequence to prevent the model from attending to the information on the subsequent positions.  
 
-### Multi-Head Attention
+### Multi-head attention
 
-The Transformer also includes a technique called **multi-head attention**.
-It allows the model to attend to different parts of the input simultaneously.
-This improves the model's ability to handle complex dependencies between input words and allows for fast processing.
+Multi-head attention allows the model to attend to different parts of the input simultaneously.
+This improves the model’s ability to handle complex dependencies between input words and allows for fast processing.
 
 ## See also
 
-- [Resources - Publications - Model architecture](https://machinetranslate.org/resources/publications/#model-architecture)
+- [Publications - Model architecture](https://machinetranslate.org/resources/publications/#model-architecture)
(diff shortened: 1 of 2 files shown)


approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
 ---
 grand_parent: Approaches
 parent: Neural machine translation
-layout: coming_soon
 title: Transformers
 description: A deep learning model architecture for neural machine translation
 ---
+
+**Transformers** are a type of neural network architecture that relies only on the [attention](/concepts/attention.md) mechanism.
```
cefoo, line 8: "**Transformers** are a type of neural network architecture that *rely* only on..."

approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
...
+It allows [neural machine translation](/approaches/neural-machine-translation.md) models to better handle the dependencies between the input and output sequences.
```
cefoo, line 9: Should we keep with the plural of `transformers` and say "They allow..."?
Or is it better to turn `transformers` into `tranformer`?

approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
...
+
+The Transformer was first introduced in a 2017 paper by Google researchers titled _"Attention Is All You Need."_
```
cefoo, line 11: I think the capitalization of `Transformer` is unnecessary.

approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
...
+
+## Architecture
+
+The Transformer consists of an encoder and a decoder.
+Each consists of several layers, which include sublayers of *self-attention* and *feed-forward neural networks*.
```
cefoo, line 16: To minimize potential quality problems when we machine translate Machine Translate, let's avoid reference words like "each". In this case, it's best to repeat the subject. :)

approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
...
+A normalisation step follows to produce the final hidden states.
+The encoder passes these hidden states to the decoder.
+
+### Decoder
+
+The decoding process is similar to the one used in the encoder, as the decoder uses output sequence embeddings as its input.
+But in addition to the self-attention and feed-forward neural network layers, the decoder uses an additional mechanism called *masked self-attention*.
+It partially masks the output sequence to prevent the model from attending to the information on the subsequent positions.  
+
+### Multi-Head Attention
```
cefoo, line 33: ### Multi-head attention

approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
...
+
+The Transformer also includes a technique called **multi-head attention**.
```
cefoo, line 35: I think we can omit this sentence.

approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
...
+It allows the model to attend to different parts of the input simultaneously.
```
cefoo, line 36: "Multi-head attention allows the model to attend to different parts of the input simultaneously."

approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
...
+This improves the model's ability to handle complex dependencies between input words and allows for fast processing.
```
cefoo, line 37: Let's use the curly apostrophe in `model’s`, following our [style guide](https://machinetranslate.org/style#apostrophes-and-quotation-marks).

approaches/transformers.md
```diff
@@ -1,7 +1,41 @@
...
+
+## See also
+
+- [Resources - Publications - Model architecture](https://machinetranslate.org/resources/publications/#model-architecture)
```
cefoo, line 41: Maybe just...?

```
- [Model architecture](https://machinetranslate.org/resources/publications/#model-architecture)
```

approaches/transformers.md
cefoo, line 11: Shall we link to the paper here?
[*Attention is All You Need*](https://arxiv.org/pdf/1706.03762.pdf)

cefoo (commented):
Hi dear @liashahnazaryan!

Thank you so much for this great PR!
I've added some comments related to formatting.

approaches/transformers.md
liashahnazaryan: I believe it's better to continue with the singular form as the original paper on this subject calls it `the Transformer architecture, with `Transformer` being the main name for the model architecture. That's why I also wrote that it's a type of neural network architecture, using singular `relies` to address the type (which you mentioned in the previous comment).

approaches/transformers.md
liashahnazaryan: I followed the format used in https://machinetranslate.org/attention. Maybe change in both then?

cefoo: Ah yeah, maybe it is better, but what do you think? I should probably add this to the style guide, as it is not mentioned.

approaches/transformers.md
cefoo: Ah perfect! Let's switch to the singular then. Should we rename the file?

liashahnazaryan: That's a good idea. I'll make the changes now.

approaches/transformers.md
liashahnazaryan: Maybe just keep:
`- [Publications - Model architecture](https://machinetranslate.org/resources/publications/#model-architecture)`
So that the readers understand that the link will take them to the `Publications` page, as in the `Attention` page, for example, the section for the relevant articles is also called `Attention mechanism` which might be misleading.

approaches/transformers.md
cefoo: > (which you mentioned in the previous comment)

I deleted it because I had interpreted it incorrectly.
I thought it was "Transformers are a type of neural network architecture that rely [the transformers]", but now I understand it's actually "Transformers are a type of neural network architecture that relies [the architecture]"...
So, my bad! Sorry about that one.

liashahnazaryan: Also, I've thought about renaming the file to `Transformer`, and I think it's best to keep it in the plural form as there can be different types of transformers with slightly changed architecture but the same idea.

approaches/transformers.md
liashahnazaryan: What do you think about this, @cefoo? :)

cefoo: Yeah, I think that's better. Thanks, Lia! :)

cefoo:
Hi @liashahnazaryan 

Adam left some comments regarding the structure for this article in issue #73.
Can you please update the article with the missing details?

Thanks so much!

liashahnazaryan:
Hey, @cefoo!
I've added a graphic from Wikipedia of the model architecture. The image is uploaded with a different PR (https://github.com/machinetranslate/machinetranslate.org/pull/416), that's why it isn't visible in this PR. 
Also, I haven't mentioned how adopted the transformer is in different MT systems, as it's widely adopted. I don't think we should mention all the companies, libraries and toolkits that use it or are based on it. Also, some of the libraries and frameworks are already included on the [`Libraries and frameworks`](https://machinetranslate.org/building/libraries-frameworks/) page. What do you think about this?

cefoo:
> Also, some of the libraries and frameworks are already included on the [Libraries and frameworks](https://machinetranslate.org/building/libraries-frameworks/) page. What do you think about this?

Sorry for the delay, Lia!
We are planning on retiring the `Libraries and frameworks` page and place the paper links in their corresponding articles.

Issue #400, opened 2023-01-17
Feature: Add alternative 


cefoo:
**Request:** Add synonyms in articles or front matter

There are topics which are named in a variety of ways. The audience may look for information using either term.

### Examples

1-
  - data privacy
  - data security
  - data confidentiality

2- 
  - machine translation
  - AI translation
  - automated translation

3-
  - quality estimation
  - quality prediction

4-
  - rule-based machine translation
  - rules-based machine translation

bittlingmayer:
In PR #426 @cefoo wrote:

> Yeah, we should add them to the article body, but perhaps also to the front matter.
Perhaps [alternateName](https://schema.org/alternateName) can be an option, as it is supported by Google Rich Results.

That sounds perfect.

cefoo:
**Note**: the `alternateName` solution doesn't work for our `jekyll-seo-tag` plugin.

PR #426, opened 2023-02-03
Add "active", "interactive" and "responsive"

bittlingmayer:
How should we deal with synonyms?

Should we cover these names, so people know that they are in the right place?

In some cases there are technical nuances, meant to be more specific or more general, but mostly these are just marketing terms.

# Description

## Type of PR

- Edits _[article name]_
- Other _[please describe]_

### Checklist:

- [ ] I have read the [contributing guidelines](contributing.md).


cefoo:
Fixes #400 

Yeah, we should add them to the article body, but perhaps also to the front matter.
Perhaps [alternateName](https://schema.org/alternateName) can be an option, as it is supported by Google Rich Results.

Issue #445, opened 2023-03-16
Hide "Supported languages" from APIs with no language support information


cefoo:
**Issue**: There are APIs that have language pages, although no language support is available.
These pages display the following message:

> The Tauyou machine translation API supports [0 languages](https://machinetranslate.org/tauyou#language-support). There are [4 integrations](https://machinetranslate.org/tauyou#integrations) with translation management systems.

See [Tauyou](https://machinetranslate.org/tauyou).

**Possible solution**: Hide the language support phrase when the `languages: [ ]` list in api.yml is empty.

bittlingmayer:
The fix is something like this: https://github.com/machinetranslate/machinetranslate.org/pull/448

bittlingmayer:
Should we just hide it?  Or do we want to be explicit?

cefoo:
> Should we just hide it? Or do we want to be explicit?

Let's have the "unknown number of languages" message as you suggested.

PR #448, opened 2023-03-16
If no supported languages listed

bittlingmayer:
# Description

Fixes the case of no supported languages listed

## Type of PR

Meta


bittlingmayer:
Fixes #445

PR #463, opened 2023-03-29
Create back-translation.md

liashahnazaryan:
# Description

Fixes # _[81](https://github.com/machinetranslate/machinetranslate.org/issues/81)_

## Type of PR

- Creates the article _[Back-translation]_

### Checklist:

### DIFF
--- a/customisation/back-translation.md
+++ b/customisation/back-translation.md
@@ -1,6 +1,18 @@
 ---
 parent: Customisation
-layout: coming_soon
 title: Back-translation
-description:
+description: Back-translating or back-copying target language sentences to augment parallel data
 ---
+
+**Back-translation** is the process of translating the monolingual data in the target language into the source language and then back into the target language.
+The goal is to generate synthetic [parallel data](/customisation/parallel-data.md) that can be used to train machine translation systems.
+
+**Back-copying** is a similar technique to back-translation.
+The process involves using an existing translation to create a new parallel sentence pair in the opposite direction.
+
+### Challenges
+
+Back-translation is challenging for language pairs with significantly different syntactic and semantic features, resulting in low-quality parallel data.
+
+Back-copying generates parallel data that is identical in the source and target languages.
+The lack of diversity can result in a machine translation system that is overly reliant on the [training data](/customisation/training-data.md) and performs poorly on unseen data.


customisation/back-translation.md
```diff
@@ -1,6 +1,18 @@
 ---
 parent: Customisation
-layout: coming_soon
 title: Back-translation
-description:
+description: Back-translating or back-copying target language sentences to augment parallel data
```
cefoo, line 4: Perhaps this definition is not clear, should we say something like "Translation of the machine translation output back into its input language"?

customisation/back-translation.md
```diff
@@ -1,6 +1,18 @@
...
 ---
+
+**Back-translation** is the process of translating the monolingual data in the target language into the source language and then back into the target language.
```
cefoo, line 7: I'd simplify "the monolingual data in the target language" to make it less wordy.

What do you think of something like this?

**Back-translation** is the process of using machine translation to translate again a machine translation output to generate synthetic data.

### Steps

1- First, the machine translation system translates a text from one language to another language.
2- Then, the system uses the machine translation output as input, and translates it back into the original language.
3- Finally, the system translates the resulting synthetic input text again into the output language.

cefoo: Do you think creating an image would make it easier to understand?

customisation/back-translation.md
```diff
@@ -1,6 +1,18 @@
...
+The goal is to generate synthetic [parallel data](/customisation/parallel-data.md) that can be used to train machine translation systems.
```
cefoo, line 8: This is a great place to introduce the "augment" term!
Perhaps we can say that synthetic data is necessary to augment parallel data so that there is more data to train machine translation systems and improve quality.

customisation/back-translation.md
```diff
@@ -1,6 +1,18 @@
...
+
+**Back-copying** is a similar technique to back-translation.
+The process involves using an existing translation to create a new parallel sentence pair in the opposite direction.
```
cefoo, line 11: I don't think it's clear how it differs with `back-translation`. Is the "existing" translation human-generated?

cefoo (commented):
Thanks so much for your hard work, @liashahnazaryan!

I hope my comments make sense. Let me know what you think. :)

customisation/back-translation.md
bittlingmayer: I disagree with this description, because it makes it sound like then it is translated from source to target.  But the training data is only translated from target to source.

Translation from source to target only happens at inference time, but that's a given.

customisation/back-translation.md
bittlingmayer: "necessary" is not always true, we can be less opinionated.

liashahnazaryan:
I'll make this a draft to work on the changes and add the suggested parts.

PR #482, opened 2023-06-16
Add Alemannic (Swiss German)

bittlingmayer:
# Description

Add Alemannic - Swiss German, Swabian, Badian, Alsatian and some Austrian dialects

https://en.wikipedia.org/wiki/Alemannic_German

## Type of PR

- Creates the article Alemannic
- Edits *TextShuttle*

### Checklist:

### DIFF
--- a/_data/apis.yml
+++ b/_data/apis.yml
@@ -787,7 +787,7 @@
    - https://www.textshuttle.ai/
    - https://www.textshuttle.ai/en/functions
   privacy_url: https://www.textshuttle.ai/en/disclaimer
-  languages: [ als, cs, da, de, en, es, fr, hu, it, ja, ko, nl, 'no', pl, pt, rm, ru, sk, sl, sv, zh ]
+  languages: [ cs, da, de, en, es, fr, gsw, hu, it, ja, ko, nl, 'no', pl, pt, rm, ru, sk, sl, sv, zh ]
   custom_languages: '*'
   more_languages_by_request: true
   glossary: true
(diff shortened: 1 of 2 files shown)


bittlingmayer:
Dutch is borderline in terms of typology, somewhere between Afrikaans and German, or English and German.

cefoo:
According to [iso639-3.sil.org](https://iso639-3.sil.org), `als` is the code for `Tosk Albanian`, and `gsw` is the code for `Alemannic`.

In their website, ModernMT mentions its support for “Tosk Albanian (als)”.
[modernmt.com/api/#languages](https://modernmt.com/api/#languages)

In this PR, we are adding `Alemannic`, but not deleting `als` from the `Albanian` entry. So if we merge this PR, two things could happen:

- Either we will add ModernMT's support for `Alemannic`, which is incorrect,
- There may be a clash, having two languages with the same ISO code.

cefoo (commented):
According to [iso639-3.sil.org](https://iso639-3.sil.org), `als` is the code for `Tosk Albanian`, and `gsw` is the code for `Alemannic`.

In their website, ModernMT mentions its support for “Tosk Albanian (als)”.
[modernmt.com/api/#languages](https://modernmt.com/api/#languages)

In this PR, we are adding `Alemannic`, but not deleting `als` from the `Albanian` entry. So if we merge this PR, two things could happen:

- Either we will add ModernMT's support for `Alemannic`, which is incorrect,
- There may be a clash, having two languages with the same ISO code.

bittlingmayer:
Thank you, that's a great catch.

`als` is used by [Alemannic Wikipedia](https://als.wikipedia.org), so I had assumed it was an ISO code of some sort, but it looks like that is just an informal, grandfathered-in Wikipedia-specific convention.

I had assumed this was a clash between ISO 639-2 and 639-3, but it seems not.

Looks like TextShuttle uses `gsw` as a base code.

cefoo:
Tested locally and works great!

PR #487, opened 2023-06-23
Merge Iranian Persian into Persian

bittlingmayer:
Persian usually refers to Iranian Persian, just like Spanish refers to Castilian Spanish.

ModernMT was being overly precise.  Other APIs just use `fa`.

# Description

`pes` is just a variant of `fa`, similar to `prs`.

## Type of PR

- Deletes the article *Iranian Persian*

### Checklist:


bittlingmayer:
I still need to re-run generate.py.

bittlingmayer:
We should check for others that only ModernMT supports, to make sure there aren't more of these.

This is similar to the PR to remove Tosk Albanian.

bittlingmayer:
We also need some way to link Persian and Tajik.

https://github.com/machinetranslate/machinetranslate.org/issues/488

PR #492, opened 2023-07-06
Remove performance bottleneck

### LINKED ISSUE #489: Bug: Build time is too long
On GitHub CI it is taking [20 minutes](https://github.com/machinetranslate/machinetranslate.org/actions/runs/5294983510/jobs/9584870628), and similarly for me locally (ThinkPad, Ubuntu).

There are some obvious performance bottlenecks, for example in language.html:
```
{% assign language = false %}
{% for lang in site.data.languages %}
  {% if lang.codes.first == page.code %}
    {% assign language = lang %}
  {% endif %}
{% endfor %}
```
So for every language, we're looping over all languages.

In general, we should try to move more out of Liquid layout templates and into generate.py.

tovmasharrison:
# Description

Fixes #489 Removed the loop asigning the language variable from language.html and handled it inside generate.py. 

### Checklist:


generate.py
```diff
@@ -212,7 +213,8 @@ def normalize(code):
     'description': f'Machine translation for { name }',
     'code': code,
     'family': family,
-    'supported_apis': supported_apis
+    'supported_apis': supported_apis,
+    'lan': language,
```
bittlingmayer, line 217: I'd call this `'info'`, since it's for the infobox.

Not good to have `language.language` or `lan.lan`.

bittlingmayer: Another option would be to just "unpack" it all into that dict.

```
    'supported_apis': supported_apis,
    **language
```

Issue #490, opened 2023-06-26
Bug: jekyll-target-blank breaks build on Ubuntu


bittlingmayer:
**Description:** On Ubuntu, .md files with non-ASCII characters break the build

**Steps to reproduce the behavior:**
```bash
bundle exec jekyll serve
```

**Screenshots [optional]:**

```bash
  Pre-Render Hooks: apis/youdao.md
  Rendering Markup: apis/youdao.md
  Rendering Layout: apis/youdao.md
     Layout source: site
jekyll 3.9.3 | Error:  URI must be ascii only "https://ai.youdao.com/DOCSIRMA/html/\u81EA\u7136\u8BED\u8A00\u7FFB\u8BD1/API\u6587\u6863/\u6587\u672C\u7FFB\u8BD1\u670D\u52A1/\u6587\u672C\u7FFB\u8BD1\u670D\u52A1-API\u6587\u6863.html"
/usr/local/lib/ruby/3.1.0/uri/rfc3986_parser.rb:21:in `split': URI must be ascii only "https://ai.youdao.com/DOCSIRMA/html/\\u81EA\\u7136\\u8BED\\u8A00\\u7FFB\\u8BD1/API\\u6587\\u6863/\\u6587\\u672C\\u7FFB\\u8BD1\\u670D\\u52A1/\\u6587\\u672C\\u7FFB\\u8BD1\\u670D\\u52A1-API\\u6587\\u6863.html" (URI::InvalidURIError)
  from /usr/local/lib/ruby/3.1.0/uri/rfc3986_parser.rb:72:in `parse'
```

**Device:**
 - Ubuntu:
 - Ruby version: 3.2.0
 - Jekyll version: 

**Additional context:**

It works on GitHub CI and on other contributors' Mac and Windows machines.

https://github.com/keithmifsud/jekyll-target-blank/issues/33

bittlingmayer:
So far, the workaround was just to disable that plugin locally, in both the _config.yml and the Gemfile.

bittlingmayer:
@cefoo Had this issue too (on Windows)

bittlingmayer:
We disabled it for now, but would be nice to add it back.

PR #494, opened 2023-07-09
Gemfile update

tovmasharrison:
# Description

Issue: When running jekyll server, I was receiving the error "bundler: failed to load command: jekyll -- require': cannot load such file -- webrick (LoadError)"

Solution: I made sure I have webrick installed by running `gem list webrick` in the terminal, if not installed run `gem install webrick` to install, and in the Gemfile I moved `gem 'webrick'` inside `group :jekyll_plugins do`.

### Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ _site
 .jekyll-cache
 .jekyll-metadata
 vendor
+.ruby-version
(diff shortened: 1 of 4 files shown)


cefoo:
Hi @tovmasharrison!
Thanks for your PR! 🚀
I am wondering if it is necessary to edit all the pages you included in this PR, I guess only a couple of files should be edited, right?
I've also noted that you've deleted `languages.html` and `generate.py`. I understand that you wanted to delete them from this PR (leave them out), but I think that, the way you did it, deletes the files from main.
Could you please check? Thank you so much!!

tovmasharrison:
I had to comment out the jekyll-target-blank because of issues with the Chinese chars in URLs (#490).

bittlingmayer:
Thanks!

PR #497, opened 2023-07-27
Suggestions to TMS PR

bittlingmayer:
- Move TMS vs CAT logic into generate.py, to keep templates in _layouts/ simpler.

- Renaming - consistency, no abbreviations
   - use `translation-management-systems` not `integrations` or `tms`
   - `api` not `engine`
   - `quality-estimation` not `quality prediction`

- Prepare for QE provider pages
   - Avoid URL clashes - QE from MT providers needs a different slug than the MT from the same provider


_layouts/translation-management-system.html
```diff
@@ -0,0 +1,72 @@
...
+</h2>
+
+<hr/>
+
+<p>
+  {% if page.api_integrations.size == 1 %}
+    The <strong>{{ page.title }}</strong> {{ page.type_description }} has <a href="#api-support">{{ page.api_integrations | size }} machine translation API integration</a>.
+  {% else %}
+    The <strong>{{ page.title }}</strong> {{ page.type_description }} has <a href="#api-support">{{ page.api_integrations | size }} machine translation API integrations</a>.
+  {% endunless %}
```
cefoo, line 20: Instead of `{% endunless %}`, it should be `{% endif %}`.

cefoo (approved):
I have tested locally and everything works great, except for the Integrations count in the landing page.
image

PR #500, opened 2023-07-28
Better SEO

bittlingmayer:
We can do similar for other pages.

# Description

Fixes the fact that queries for "list of machine translation apis" doesn't find us

Ideally we'd also find a way to "engines".

## Type of PR

- Edits _APIs_

### Checklist:


cefoo (approved):

bittlingmayer:
Can you take another look - I added the rest.

PR #502, opened 2023-07-28
Link articles to community

bittlingmayer:
Links every article to the community

For example, at the bottom of the ModernMT article:

[**Search the community for *ModernMT***](https://www.reddit.com/r/machinetranslation/search?q=ModernMT&restrict_sr=on)


cefoo (approved):
This is great!

I checked in a local copy and everything is functional.
However, I would put each button in a different line, as it currently looks like this:
imagen

What do you think, @bittlingmayer?

PR #506, opened 2023-07-30
Add jekyll-include-cache

bittlingmayer:
# Description

Potentially partly address #489

## Type of PR

- Meta / infra

### Checklist:


bittlingmayer:
We don't actually have very many `include` statements - it would help more if done for the Just The Docs nav, but that's not in our code.

PR #520, opened 2023-08-06
Collapsible lists in languages, APIs etc

bittlingmayer:
# Description

These lists are getting really long, and if there are multiple, the lower ones are hard to discover.
 
## Type of PR

- Meta / infra

### Checklist:

### DIFF
--- a/_layouts/aggregator.html
+++ b/_layouts/aggregator.html
@@ -39,9 +39,9 @@ parent: Aggregators
   <details>
     <summary>
       {% if page.supported_apis.size == 1 %}
-        <p><strong>1</strong> API supports {{ page.title }}.</p>
+        <strong>1</strong> API supports {{ page.title }}.
       {% else %}
-        <p><strong>{{ page.supported_apis | size }}</strong> APIs support {{ page.title }}.</p>
+        <strong>{{ page.supported_apis | size }}</strong> APIs support {{ page.title }}.
       {% endif %}
       <p class="preview hint">
         {{ page.supported_apis | slice: 0, 5 | map: 'name' | join: ', ' }}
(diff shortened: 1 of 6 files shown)


cefoo:
This works perfectly for the left menu in Languages.
In the body, the list is expanded by default. (Is this on purpose?)

image

When shrinked, `Baidu Translate` is visible in a small font.

image

image

But when a language is just supported by `Baidu Translate`, it looks fine.

image

cefoo (commented):
Overall, it looks great! We just need to fix that `Baidu Translate` issue.

bittlingmayer:
> In the body, the list is expanded by default. (Is this on purpose?)

No, sorry - nasty unintuitive behaviour!

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#attributes

> **Note:** You have to remove this attribute entirely to make the details hidden. `open="false"` makes the details visible because this attribute is Boolean.

Fixed

bittlingmayer:
> When shrinked, Baidu Translate is visible in a small font.

I did this on purpose.  I want to leave a preview of 5 or so, and an ellipses if necessary, to make it more discoverable.

Else I worry that people won't click on the twirl down or summary, and won't notice that they can expand.

Since the preview is wrapping horizontally and in small font, it can even be 10 or 20.  The point is not to have 300.

bittlingmayer:
I can't understand why it only showed 1 in the preview.

cefoo:
> I can't understand why it only showed 1 in the preview.

I fear that if it's only one, it's not clear that it is a preview, it just looks like an issue to me.
But it is a list, then it will probably be clearer.

I'm testing locally after the last changes.

cefoo:
image

cefoo:
Looks great!

image

PR #530, opened 2023-08-20
Improve build time by caching

### LINKED ISSUE #489: Bug: Build time is too long

tovmasharrison:
# Description

Fixes #489 I have updated just-the-docs remote theme version to the latest and implemented a caching mechanism with jekyll-include-cache plugin to the the included sidebar.html and icons.html inside default.html by overriding default.html from the remote theme. Also, I have cached the social media includes inside footer_custom.html.

Additonally, I have enabled incremental builds and set lsi: false, as there is no generation of similiar posts implemented.

The above steps have reduced the build time from 2779 seconds to 342 seconds.

### Checklist:

### DIFF
--- a/_includes/components/sidebar.html
+++ b/_includes/components/sidebar.html
@@ -1,0 +1,82 @@
+{%- comment -%}
+  Include as: {%- include components/sidebar.html -%}
+  Depends on: page(?), site.
+  Results in: HTML for the side bar.
+  Includes:
+    title.html, nav.html, nav_footer_custom.html
+  Overwrites: 
+    pages_top_size, collections_size, collection_entry,
+    collection_key, collection_value, collection, nav_footer_custom.
+{%- endcomment -%}
+
+<div class="side-bar">
+  <div class="site-header" role="banner">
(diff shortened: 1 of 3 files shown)


bittlingmayer:
Looks good to me.  I think we should add some comments in the file and in the README explaining where it is from.

bittlingmayer:
Awesome!  Build time is down from 40 minutes to [6 minutes](https://github.com/machinetranslate/machinetranslate.org/actions/runs/6035205551/job/16375117225
)

PR #541, opened 2023-09-26
Add memoQ quality estimation integrations

bittlingmayer:
# Description

memoQ has integrated both ModelFront and TAUS.

## Type of PR

- Add integrations

### Checklist:


bittlingmayer:
@tovmasharrison if you approve of the changes, maybe you can take over and run generate.py?

tovmasharrison:
@bittlingmayer Looks good to me.

_layouts/quality_estimation.html
```diff
@@ -12,19 +12,17 @@ <h2>
...
 
 <p>
-  {% if page.supported_languages.size > 0 %}
-    The <strong>{{ page.title }}</strong> quality estimation API supports <a href="#language-support">{{ page.supported_languages | size }} languages</a>.
-  {% else %}
-    The <strong>{{ page.title }}</strong> quality estimation API supports an unknown number of languages.
-  {% endif %}
+  The <strong>{{ page.title }}</strong> quality estimation API is available in
+  <a href="#integrations">{{ page.integrations | size }}
+  machine translation management system{% if page.integrations.size == 1 %}s{% endif %}</a>.
```
cefoo, line 17: Can we have different versions depending on the QE API integrations count? Since there aren't a lot of integrations yet, only 0, 1 or 2, I think singular and plural sentences are too visible.

Screenshot 2023-09-26 at 11 04 28

cefoo (commented):
A couple of comments:

-  With `{{ page.description }}` in `integrations.html`, Integration pages titles are “Machine translation API integrations in Trados Studio”. Should we change that? Does this properly describe QE tools as well?
- I don’t think the way we display the tagline is clear. Plus, imo, the format could be improved.
Screenshot 2023-09-26 at 11 06 46

  - **Clarity:** From where it is located, it looks like a subtitle or a quote, but it’s not really related to what comes afterward. Perhaps we can place it before the body? Or just something like `For Omniscien, quality estimation is called translation confidence scoring and quality estimates` (or something like that).
  - **Format:** I don’t like the off-center alignment; it seems out of place. Regardless, I'm concerned that giving the tagline too much visual prominence might come across as overly promotional.

- Can we make `Customization` lowercase?
Screenshot 2023-09-26 at 11 20 27

- Should we make the list introduction shorter? Something like `1 translation management system integration` or similar.

Screenshot 2023-09-26 at 11 24 08

bittlingmayer:
> With {{ page.description }} in integrations.html, Integration pages titles are “Machine translation API integrations in Trados Studio”. Should we change that? Does this properly describe QE tools as well?

I'd say it broadly does, I mean, quality estimation is very much a part of machine translation, in terms of its origins, research, technology and how the human translation industry thinks of it.

If there is an observable shift to something like "AI API integrations in Trados Studio" then we could follow that, but I wouldn't want to lead it.

If there is a partial shift, then we could do something like "Machine translation and artificial intelligence API integrations in Trados Studio" but so far we have avoided "AI" and "artificial intelligence". :-)

_layouts/quality_estimation.html
bittlingmayer: I fixed this, should be `!= 1` not `== 1`.

Once we had connectors etc that already exist (a bit like aggregators for MT), it will change to 5 or so for ModelFront and I can only assume that the others will launch more too.  I just didn't want to do that in this PR.

There will be a few like Phrase MTQE or KantanQES that are only available in 1 and for that I think it is useful to be explicit, it is a bit like the Crowdin MT, i.e. that property of being very tied to that TMS is very important.

I'm also trying to keep us from having too much repeated code for plural forms, both here and in apis.html etc.  I know I added it originally, but I regret it - more code, hard to maintain, all for questionable benefit to the user.

bittlingmayer:
> Regardless, I'm concerned that giving the tagline too much visual prominence might come across as overly promotional.

Yes, I had not noticed this.  h3 was not right.  I made it normal text and in quotation marks.  These are not really marketing taglines but just their description of the technology their providing.  Since the category is still evolving.

The real reason I lean towards ripping this out is that it's not evergreen.  For QE alone it seems we could handle maintaining it but to handle it for all the MT APIs doesn't scale, since unlike languages, it's hard to notice when it changes.

bittlingmayer:
> Should we make the list introduction shorter? Something like 1 translation management system integration or similar.

Done

PR #546, opened 2023-10-10
Add the codes for unlisted languages to languages list

### LINKED ISSUE #486: Feature: Languages article should list languages for which we only have a code.
https://machinetranslate.org/languages lists only those 226 languages for which we have an entry in [languages.yml](https://github.com/machinetranslate/machinetranslate.org/blob/master/_data/languages.yml).

In reality, [apis.yaml](https://github.com/machinetranslate/machinetranslate.org/blob/master/_data/apis.yml) lists about 400 codes.

In the *Languages* article, we could just list codes for the rest, like we do on the individual API pages, like [Niutrans](https://machinetranslate.org/niutrans).

That way, we could give a total number, both in that article, and on the home page.

tovmasharrison:
# Description

In the Languages article, I have just listed codes for the languages for which we don't have an entry. 

Fixes #486 

### Checklist:


cefoo:
Hi @tovmasharrison!

Thank you for your PR!
I have tested it locally and it is fully functional! Thanks!

This is a personal impression, but I feel like, for languages we just have codes, perhaps it would be neater to not display the code twice.

When this PR is merged, we would see the language list (this is taken from `niutrans.md`).
Screenshot 2023-10-11 at 09 38 31

As Adam mentioned in #486, it would be great to automatically get the languages names for those codes. However, I'm not sure how necessary it is, since as soon as we create language pages for each of these codes, then we won't need that solution anymore… Unless an API starts supporting languages that aren't already supported by Machine Translate. (For another issue).

tovmasharrison:
Hi @cefoo!

No problem! 

Just let me know if only the language code needs to be displayed.

bittlingmayer:
Re `crp` @cefoo that code is just for Creoles in general - so that's some Lilt-specific code, which we need to add to https://github.com/machinetranslate/machinetranslate.org/blob/master/_data/api_language.json to redirect to its proper code.

bittlingmayer:
We don't actually need to put this in the .md files:
Screenshot 2023-10-21 at 23 49 18

In generate.py, we could delete `supported_language_count` after we use it to sort.

bittlingmayer:
We need to update languages.html template, so that if we have a code but no name:
- We should put it in backticks, at least in the article text, and ideally also in the title and description.
- We should add a note:

> This language has not yet been fully added to Machine Translate.
>
> You can add it to Machine Translate by getting the name, language family, script and typology to [languages.json](https://github.com/machinetranslate/machinetranslate.org/blob/a2188c58f1389edbd62c6b77ec8146adcebc4049/_data/languages.json) and submitting a pull request.
>
>  You may be able to find this information in [its Wikipedia article](https://en.wikipedia.org/wiki/ISO_639:gom).

bittlingmayer:
> In their documentation, Lilt just mentions "Creole" as a supported language: 
> What can we do in that case?

I looked through their docs, but they don't list it anywhere.

Write to their support?  (Or sales - they'll answer faster. ;-))

My guess would be Haitian Creole, but it could be some other French-, English-, Spanish- or Portuguese-based Creole.

bittlingmayer:
I added a bug here: https://github.com/machinetranslate/machinetranslate.org/issues/553

bittlingmayer (approved):
I'll merge once conflicts are resolved.

I created a bug about Lilt's "Creole": https://github.com/machinetranslate/machinetranslate.org/issues/553

tovmasharrison:
Conflicts are resolved. (cc) @bittlingmayer

cefoo:
> > In their documentation, Lilt just mentions "Creole" as a supported language:
> > What can we do in that case?
> 
> I looked through their docs, but they don't list it anywhere.
> 
> Write to their support? (Or sales - they'll answer faster. ;-))
> 
> My guess would be Haitian Creole, but it could be some other French-, English-, Spanish- or Portuguese-based Creole.

I'll ask!

PR #556, opened 2023-10-24
Update Machine Translate to MTF

cefoo:
# Description

Changes mentions of `Machine Translate` to `Machine Translate Foundation`

# Comment

I haven't changed every instance of `Machine Translate`.
I believe that we should keep using `Machine Translate` when talking specifically about the repo and the website. To me, the `Machine Translate Foundation` should be used to refer to the organization itself.

Another comment -- should we capitalize `Foundation`? I did, I like it, but I can revert it.

## Type of PR

- Edits certain articles where Machine Translate is mentioned

### Checklist:


_config.yml
```diff
@@ -4,7 +4,7 @@
 domain: machinetranslate.org
 url: https://machinetranslate.org
 title: Machine Translate
-tagline: Open information and community for machine translation
+tagline: The Machine Translate Foundation
```
bittlingmayer, line 7: I think it still needs an actual tagline.

about.md
```diff
@@ -16,7 +16,7 @@ seo:
 
 ## Mission
 
-Machine Translate’s mission is to **make machine translation more accessible to more people** as the field of machine translation grows.
+The Machine Translate foundation’s mission is to **make machine translation more accessible to more people** as the field of machine translation grows.
```
bittlingmayer, line 19: Uppercase, i'd say

about.md
```diff
@@ -89,7 +89,7 @@ The society is registered in Zug, Switzerland and is run as a non-profit.
 
 Machine Translate is created and edited by [contributors](/contributing/contributors.md) from all over the world.
 
-Machine Translate is also supported by the generosity of other organizations and by people like you!
+The Machine Translate foundation is also supported by the generosity of other organizations and by people like you!
```
bittlingmayer, line 92: ditto

about.md
```diff
@@ -98,7 +98,7 @@ Machine Translate is also supported by the generosity of other organizations and
 ## Contact
 To contact us about content, you can use the normal open channels for [contributing](/contributing/contributing.md).
 
-To contact us about the foundation and supporting Machine Translate, you can email [contact@machinetranslate.org](mailto:contact@machinetranslate.org).
+To contact us about the foundation and supporting the Machine Translate foundation, you can email [contact@machinetranslate.org](mailto:contact@machinetranslate.org).
```
bittlingmayer, line 101: ditto

bittlingmayer:
> I haven't changed every instance of Machine Translate.  I believe that we should keep using Machine Translate when talking specifically about the repo and the website.

I'm not sure.  But I'd be fine with it, to make this change faster, and then we see how it looks.

_config.yml
bittlingmayer: Just from an SEO perspective...

bittlingmayer: It'll show up as:
> Machine Translate | The Machine Translate ...
which is not a very good use of real estate.

This is what is currently shown:
Screenshot 2023-10-27 at 00 56 23

3 options that occur to me:
1. Add "... Foundation" to the title, keep the tagline as it was
2. Add "Foundation for... " to the tagline, keep the title as it was
3. Keep both as they were

I lean towards 1.

cefoo: Yeah, that's true, I agree. I've fixed it.

PR #562, opened 2023-10-30
Change aggregator wording

bittlingmayer:
# Description

The wording was backwards.  The APIs don't support the aggregator, the aggregator supports the APIs.

## Type of PR

- Edits all the generated articles in /aggregators

### Checklist:

### DIFF
--- a/_layouts/aggregator.html
+++ b/_layouts/aggregator.html
@@ -54,7 +54,7 @@ parent: Aggregators
       {% if page.supported_apis.size == 1 %}
         <strong>1</strong> machine translation API supports {{ page.title }}.
       {% else %}
-        <strong>{{ page.supported_apis | size }}</strong> machine translation APIs support {{ page.title }}.
+        <strong>{{ page.supported_apis | size }}</strong> machine translation APIs are supported by the {{ page.title }} aggregator.
       {% endif %}
       <p class="preview hint">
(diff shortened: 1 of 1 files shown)


_layouts/aggregator.html
```diff
@@ -54,7 +54,7 @@ <h2>Machine translation</h2>
       {% if page.supported_apis.size == 1 %}
         <strong>1</strong> machine translation API supports {{ page.title }}.
       {% else %}
-        <strong>{{ page.supported_apis | size }}</strong> machine translation APIs support {{ page.title }}.
+        <strong>{{ page.supported_apis | size }}</strong> machine translation APIs are supported by the {{ page.title }} aggregator.
```
cefoo, line 57: To honor the "no passive voice" instruction, could we change it to... ?

`The {{ page.title }} aggregator supports <strong>{{ page.supported_apis | size }}</strong> machine translation APIs.`

Does it sound more natural with the passive voice?

cefoo (approved):

_layouts/aggregator.html
bittlingmayer: I think my idea was to have the bold part in front.  But I can change that.

bittlingmayer: I think the prohibition on passive voice was more in sentences without "by".  e.g. "SMT is considered the best approach to machine translation."  By whom, exactly?

PR #584, opened 2023-11-28
Add Sentence embedding to roadmap, and create MT Summit page

tovmasharrison:
# Description
Sentence embedding was hyperlinked, but we didn't have a page for it, so I added it to the roadmap. 

Also, I created a page for MT Summit to include all of its events, but I think more information needs to be represented on that page. 

### Checklist:


cefoo:
Hi @tovmasharrison!

Thanks for working on this PR!

The thing with the MT Summit is that it is not an association. It is an event that is organized by EAMT, AAMT, and AMTA. It takes place bianniely, and those associations take turn in organizing it. The last MT Summit was this year, organized by AAMT. EAMT will be in charge of the next one in 2025.

I think it's great to have a filtered list with MT Summit events, but perhaps that list should be in `events.md`?
Let me know what you think!

Also, perhaps we could list the corresponding MT Summit editions in AMTA's, EAMT's and AAMT's pages (the ones each of them organized). What do you think?

more/associations/mt-summit.md
```diff
@@ -0,0 +1,23 @@
+---
+grand_parent: More
+parent: Associations
```
cefoo, line 3: Maybe we can do...

`parent: Events`

...and then, in `events.md`, add a section before the first table in which we have filtered content, like `AMTA`, `MT Summit`, `EAMT`, `WMT`, and `AAMT`?

tovmasharrison:
Hi @cefoo !

I have deleted the separate page for Mt Summit and included its events as a list inside ```events.md```. Hence, whenever we need to link MT Summit to show all of its events, we can use its hash (e.g. /events#mt-summit).

Let me know what you think.

cefoo:
Looking good!

What do you think, @bittlingmayer?
Screenshot 2023-12-18 at 08 32 37

Screenshot 2023-12-18 at 08 32 26

cefoo:
Hi @tovmasharrison!
Do you think we could create AMTA and EAMT sections like the one for MT Summit on top of the events table? Perhaps we could do it in a way that it looks like buttons or tabs, so that it doesn't take up much visual space? (Just thinking out loud). Thanks!

tovmasharrison:
> Hi @tovmasharrison!
> Do you think we could create AMTA and EAMT sections like the one for MT Summit on top of the events table? Perhaps we could do it in a way that it looks like buttons or tabs, so that it doesn't take up much visual space? (Just thinking out loud). Thanks!

Hi @cefoo ,

Yes, I can create sections for AMTA and EAMT similar to MT Summit.

No problem!

bittlingmayer (approved):

PR #591, opened 2023-12-31
Adding results to wmt23.md

liashahnazaryan:
# Description

Adding results to WMT23. 

## Type of PR

- Edits _[WMT23]_

### Checklist:


events/wmt23.md
```diff
@@ -198,43 +239,45 @@ In 2022, the *News* machine translation task was renamed the *General* machine t
...
-### Research Papers
-
-> Research papers should describe original research corresponding to the categories listed above. Research papers that have been or will be submitted to other meetings or publications must indicate this at submission time, and must be withdrawn from the other venues if accepted and published at WMT 2022.
-
-> We will not accept for publication papers that overlap significantly in content or results with papers that have been or will be published elsewhere. It is acceptable to submit work that has been made available as a technical report (or similar, e.g. in arXiv) without citing it.
-
-> For the research track, papers should be anonymised, be between 6 and 10 pages in length (excluding references) and may include supplementary material.
+## Results
 
+Full results of the shared tasks:
```
liashahnazaryan, line 244: Hey @cefoo! I've noticed we keep talking about "Full results of the shared tasks" here, but the link we share is only for the news/general translation task. In the past (excluding 2022), the file we refer to used to cover not just the news/general translation task but also a bunch of other translation tasks in the findings. However, starting from 2022, each shared task's results are in separate files.
Here's my suggestion: we could either share links to all shared task results like I did here and also update all the previous year pages accordingly. Or, we could just jump straight into the news/general task and provide the link to the full results for that shared task. I'm leaning towards the latter option since we mostly concentrate on the news/general task. What do you think?

cefoo: Hi @liashahnazaryan!

Thank you so much for your PR!

Yes, I agree with you. We mainly focus on the General/News task, so let's keep sharing those results, and link the rest, as you suggested.

liashahnazaryan: Dear @cefoo, I've updated all the articles, where the results file includes only the general/news task. For WMT06-17, the file includes results for all shared tasks, so I left the articles for those years unchanged.

events/wmt23.md
```diff
@@ -61,6 +61,47 @@ It was organised by [WMT](/wmt).
...
+
+
+### System Papers
+
+> System papers must describe one or more shared task submissions. System paper submissions that we cannot link to a shared task submission will be rejected without review. System papers can overlap with other published work, and do not have to follow the double submission policy. There is no maximum length for system papers, but normally a short paper (4-6 pages) is appropriate. System papers should not be anonymised.
+
+
+## Poster format
+
+- System description papers will be presentated as posters.
```
bittlingmayer, line 95: Do we even need this?

It's not evergreen because of the verb tense, so if we keep it, I think we should just use a quote, or make it past tense.

events/wmt23.md
```diff
@@ -61,6 +61,47 @@ It was organised by [WMT](/wmt).
...
+- Poster panels are 3.28 foot (1 meter wide) x 8.20 foot (2.5 meter tall). Portrait orientation is suggested.
```
bittlingmayer, line 96: Ditto

bittlingmayer (approved):
Looks good, just one small comment

events/wmt23.md
```diff
@@ -198,43 +233,34 @@ In 2022, the *News* machine translation task was renamed the *General* machine t
...
+#### English →
 
+| Language pair | System | Average score |
+| --- | --- | --- |
+| → German | `GPT4-5shot` | 89.0 |
+| → Czech | `ONLINE-W` | 84.1 |
+| → Chinese | `Yishu`| 82.2 |
+| → Japanese | `GPT4-5shot` | 79.5  |
 
-## Poster format
```
liashahnazaryan, line 230: I recommend removing this entire subsection, @bittlingmayer and @cefoo, as it contains very specific information that may not be relevant to the readers.

events/wmt22.md
```diff
@@ -125,11 +125,6 @@ Confirmed language pairs:
 > System papers must describe one or more shared task submissions. System paper submissions that we cannot link to a shared task submission will be rejected without review. System papers can overlap with other published work, and do not have to follow the double submission policy. There is no maximum length for system papers, but normally a short paper (4-6 pages) is appropriate. System papers should not be anonymised.
 
 
-## Poster format
-
-Posters must be presented in Gather.Town.
```
liashahnazaryan, line 130: Have done the same here, as this one doesn't even follow the rules in the [website](https://www.statmt.org/wmt22/#:~:text=PRESENTATION%20AND%20POSTER%20FORMAT&text=For%20the%20in%2Dperson%20poster,that%20are%20portrait%20in%20orientation.).

cefoo: Yeah, that seems to changed from the time I included it. Thanks!

PR #598, opened 2024-02-11
Fix slug description for 3 letter language codes

tovmasharrison:
# Description

There was a bug where ```None``` was generated as the name inside the SEO description for languages we only have codes. Now, instead of ```None```, the language code is being generated inside the description as the name. 

### Checklist:


generate.py
```diff
@@ -427,6 +427,7 @@ def normalize(code):
     slug = code
     frontmatter['title'] = f'<code>{code}</code>'
     frontmatter['description'] = f'Machine translation for <code>{code}</code>'
+    frontmatter['seo']['name'] = f'Machine translation for {code}'
```
bittlingmayer, line 430: I think we can simply this a lot:

```
  frontmatter = {
    'autogenerated': True,
    'autogenerated_from': 'languages.json',
    'nav_order': 1000 - len(supported_apis),
    'nav_exclude': True,
    'parent': 'Languages',
    'layout': 'language',
    'title': name or f'`{ code }`',
    'description': f'Machine translation for { name or f'`{ code }`' }',
    **language,
    'family': family,
    'supported_apis': supported_apis,
    'supported_qe_apis': supported_qe_apis,
    'seo': {
      'name': f'Machine translation for { name or code }',
      'type': 'Language'
    }
  }
  
  slug = slugify(name) if name else code
```

Not sure if we can use \` or if we need to use `<code>`...`</code>`.

PR #621, opened 2024-05-14
Fix event date generation

tovmasharrison:
# Description

Improved the way the start and end dates were generated for events by including the month in the start date just like it is included in the end date in the opening sentence to avoid the below 👇 

### Checklist:


cefoo:
Thanks, Tom!!

PR #641, opened 2025-07-26
Add Input-types section

tovmasharrison:
# Description

Added support flags for text and voice input for each API.
These flags will be used to generate a new `Input Types` section in the sidebar via `generate.py`, allowing users to view input types (text, voice) along with the list of APIs that support each.

### Checklist:


_layouts/api.html
```diff
@@ -76,6 +103,15 @@ <h2>
     <dd><a href="#integrations">{{ page.integrations | size }} translation management systems</a></dd>
   {% endif %}
 
+  {% if page.inputs.size > 0 %}
+    <dt>Input Type</dt>
```
bittlingmayer, line 107: Input modes

tovmasharrison: I also changed the section/page name to Input modes as well.

_data/apis.json
```diff
@@ -7824,6 +7964,10 @@
                 "ja"
             ]
         ],
+        "inputs": [
+            "speech",
```
bittlingmayer, line 7968: afaik this is false - it supports files, but not speech

bittlingmayer (changes requested):
we gotta be really sure about the input modes

i don't think there is any shortcut for going and checking manually...

input-modes/text/text.md
```diff
@@ -0,0 +1,24 @@
+---
+parent: Input modes
+layout: coming_soon
```
bittlingmayer, line 3: we can remove this now

input-modes/speech/speech.md
```diff
@@ -0,0 +1,24 @@
+---
+parent: Input modes
+layout: coming_soon
+nav_order: 2
+has_toc: false
+title: Speech
+description: Speech-to-speech machine translation, text-to-speech machine translation, speech-to-text machine translation
```
bittlingmayer, line 7: Speech input mode for machine translation APIs

PR #638, opened 2025-07-09
Add GenAI apis that are also used for translation

tovmasharrison:
# Description

Added 7 GenAI apis that are also used for translation. Additionally, I needed to do a quick syntax fix on `generate.py`, and add some languages to `languages.json`.

### Checklist:

- [x] I have read the [contributing guidelines](/CONTRIBUTING).
- [x] I have followed the [style guide](http://machinetranslate.org/style).

PR #642, opened 2025-07-30
Add tags to distinguish between MT and GenAI

tovmasharrison:
# Description

Added tags in the `apis.json` file for each API indicating whether it's MT, GenAI or both. Now, on each API's page, it shows the indication in the blockquote. 

### Checklist:


_layouts/api.html
```diff
@@ -105,11 +105,29 @@ <h2>
     </ul>
   {% endif %}
 
-  {% if page.self_serve == false %}
     <hr/>
+  {% if page.type %}
+    {% if page.type == 'mt' %}
+      <p>🌐 MT</p>
```
bittlingmayer, line 111: - No emojis
- No acronyms
- No title case

tovmasharrison:
> After thinking, I think that "type" is a bit misleading, and we shouldn't use the word "genAI".
> 
> We can just not track the underlying implementation, same way that we don't worry about SMT vs NMT, because it's constantly changing and also too often more of a marketing thing.
> 
> Really we just want to warn if prompting is required. (Which is 1:1 with `gen_ai`, no?)
> 
> Because that's actually concrete and outward-facing.

@bittlingmayer Yes, I agree! If prompt not required, then it's not `gen_ai`. Hence, it is translation specific, and vice versa.

_layouts/api.html
```diff
@@ -34,24 +34,33 @@ <h2>
...
-  {% if page.customisation contains 'Adaptive' %}
-    The API also supports <a href="/adaptive">adaptive</a> customisation.
-  {% endif %} <br>
-  {% if page.file_translation %}
-    The API also supports <a href='/file-translation'>file translation</a>.
+
+<p>
+  {% if page.prompt_required == true %}
+    The <strong>{{ page.title }}</strong> API requires a prompt for translation.
+  {% elsif page.prompt_required == false %}
```
bittlingmayer, line 62: i think we don't need to do anything in this case

_layouts/api.html
```diff
@@ -34,24 +34,33 @@ <h2>
...
-  {% endif %}
-  {% if page.customisation contains 'Adaptive' %}
-    The API also supports <a href="/adaptive">adaptive</a> customisation.
-  {% endif %} <br>
-  {% if page.file_translation %}
-    The API also supports <a href='/file-translation'>file translation</a>.
+
+<p>
+  {% if page.prompt_required == true %}
+    The <strong>{{ page.title }}</strong> API requires a prompt for translation.
```
bittlingmayer, line 61: The <strong>{{ page.title }}</strong> API is <strong>not translation-specific</a>, so for translation it <strong>requires a prompt</a>.

_data/apis.json
```diff
@@ -406,7 +406,7 @@
             "https://www.acolad.com/en/technology/machine-translation.html"
         ],
         "languages": [],
-        "self-serve": false
+        "self-serve": true
```
bittlingmayer, line 409: I think it's still false, no?

On their page it says "Get in touch", it doesn't just let you try it out.

_layouts/api.html
```diff
@@ -105,11 +114,17 @@ <h2>
     </ul>
   {% endif %}
 
-  {% if page.self_serve == false %}
     <hr/>
+
+  {% if page.prompt_required %}
+    <p>Requires prompt</p>
+  {% else %}
+    <p>No prompt required</p>
```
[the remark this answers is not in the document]
tovmasharrison: Removed

tovmasharrison:
> https://www.acolad.com/en/technology/machine-translation.html

@bittlingmayer Fixed

tovmasharrison:
@bittlingmayer 

All done. 

Also done:
- https://github.com/machinetranslate/machinetranslate.org/pull/638#discussion_r2265733592
- https://github.com/machinetranslate/machinetranslate.org/pull/638#discussion_r2265737401

PR #645, opened 2025-08-14
Rename aggregators to routers

tovmasharrison:
# Description

Renamed aggregators to routers, since this is now the established term in the AI world.

### Checklist:


routers/routers.md
bittlingmayer, line 1: To keep the history, this should be `git mv`'ed from aggregators.md, not deleted and added.

bittlingmayer (approved):

PR #647, opened 2025-08-22
Update models.md

cefoo:
# Description

Updates [models.md](https://machinetranslate.org/models)

### Checklist:

### DIFF
--- a/models/models.md
+++ b/models/models.md
@@ -5,12 +5,30 @@ title: Models
 description: Large language models
 ---
 
-**Large language models** (**LLMs**) are a type of generative AI (GenAI) trained on very large collections of text. They learn patterns and can generate text that is fluent and coherent.
+Machine translation **models** are machine learning models for generating a translation.
 
-LLMs are general-purpose. They can translate, summarize or answer questions when given instructions. These instructions are called *prompts*.
+[APIs](/apis) provide convenient access to underlying models. Models can also be trained and deployed directly.
 
-Unlike traditional machine translation, like [neural machine translation](/neural-machine-translation), they are not trained only for translation.
+## Evolution
 
-LLMs can use context because they pay attention to the sentences before and after the input to give more natural results.
+With [statistical machine translation](/statistical-machine-translation), systems required multiple models:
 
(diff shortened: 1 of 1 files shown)


bittlingmayer:
I think this text would fit more in an article "GenAI for translation" or something like that.

This is just about our Models section, which includes many different types of models, not just genAI models.
- neural MT models
- translation-focused genAI
- generic genAI
- other translation-adjacent models, like QE and APE
...

(Also, re "LLMs", I prefer to leave it defined as more than genAI/chat models.  So it includes seq2seq models like T5, and in the wider sense, masking models like BERT.  I don't want to fight popular usage Schmidhuber style, but I also don't want us to constantly retroactively redefine things.  Concretely, if a model is an "LLM" today, then in 5 years it is still an LLM, regardless of what happens between now and then.)

So I think we just want to define what a model is, and how it relates to APIs, model families vs model versions, what the relevant types are i.e. what we cover.  (Followed by the list of models)

> Machine translation **models** are machine learning models for generating a translation.
>
> [APIs](/apis) provide convenient access to underlying models.  Models can also be trained and deployed directly.
>
> ## Evolution
> With [statistical machine translation](/statistical-machine-translation), systems required multiple models.  With [neural machine translation](/neural-machine-translation), that was replaced by one end-to-end model.  Increasingly, one model is used for many language pairs.
>
> Unlike machine translation APIs or apps, raw machine translation models typically still do not handle basic requirements of a production system, like [language identification](/language-identification), [segmentation](/segmentation) or [bridging](/bridging).
>
> Each model typically consists of multiple model versions, that vary by size or release date.
> 
> ## More types
> In addition to neural machine translation models, there are various types of models currently used by machine translation APIs and related APIs.
> - Generic genAI (chat) models, that require [prompting](/prompting) to generate translations
> - Translation-focused genAI (chat) models
> - [Quality evaluation](/quality-evaluation) models
> - [Quality estimation](/quality-evaluation) models
> - [Automatic post-editing](/automatic-post-editing) models
>

bittlingmayer (approved):

PR #648, opened 2025-08-23
Autogenerate companies

tovmasharrison:
# Description
- Autogenerated **Companies** page
- Added missing companies in `companies.json`, and for each company, added data specifying their offerings, which then gets mentioned on each company page. 
- Changed all potential duplicate API, QE, APE, TMS/CAT, and router IDs, and added an extension based on their type. 

### Checklist:

### DIFF
--- a/.github/tests/test_slugs.py
+++ b/.github/tests/test_slugs.py
@@ -15,10 +15,16 @@ def test_unique_slugs():
 
     url_elements = soup.find_all('loc')
 
+    skip_slugs = ['integrations', 'routers', 'quality-estimation', 'automatic-post-editing', 'models']
     slugs = []
     for url_element in url_elements:
         url_text = url_element.text.split('/')[-1]
         slug = url_text
+        
+        # Skip mentioned dirs in skip_slugs
+        path_segments = url_element.text.split('/')[3:]
+        if any(segment in skip_slugs for segment in path_segments):
+            continue
 
         # Check duplicate/conflicting slugs
         assert slug not in slugs, f'Duplicate Slug {slug} for the URL {url_element.text}: {slug} or unnecessary trailing slash'
@@ -26,6 +32,9 @@ def test_unique_slugs():
 
         # Check path without 'files' prefix
         check_path = url_element.text.split('/')[-2]
+        
+        if check_path in skip_slugs:
+            continue
 
         # Check that URL is one level deep
         if check_path != 'files':


cefoo:
We have companies information at https://machinetranslate.org/companies that we should perhaps consider to leave all info from the same company under the same <id> page.
You said "No company pages", but do you mean no paragraph-like information about the company?
I don't like lots of subpages with different ids for the services offered by the same company, but we sort of do that for organizations and events (eg /amta, /amta-2022, /amta-2023, etc).

cefoo:
I think it'd be neater to have a centralized place where you can get all services offered by the same company, specially since some companies are an engine + TMS, other are just engines, others are QE + APE, other are connectors, etc.

However, making sure we have proper links would work too. For example, at the bottom of the QE article, we have a list of the QE products offered by companies. If we click on those, we get to a dedicated product page. I think we should add a link there to "other products offered by the same company". That way, for example, we'd have a linked path like: QE article -> ModelFront QE page -> ModelFront APE page.

bittlingmayer:
The links (at the bottoms of product pages, and in Companies) seem like the easy way to move forward today.

An actual centralized page will require more thought.

Currently our overall framing is:

> Quality estimation > ModelFront  
> Quality estimation > Google  
> ...

Basically it is from the user's perspective.  It is useful if a user is looking for a specific technology.

Company pages would push us towards:

> ModelFront > Quality estimation  
> ModelFront > Automatic post-editing  
> ...

That would make it more about the companies.

tovmasharrison:
> By the way @cefoo
> 
> What do you think about adding a URL subdir for anything not MT?
> 
> * `/google-translation-hub-mtqp` --> `/quality-estimation/google
> * `/modelfront` --> `/quality-estimation/modelfront`
> * `/modelfront-ape` --> `/automatic-post-editing/modelfront`
> 
> Basically remove the redirects for those, go back to the default behaviour of Jekyll.
> 
> But leave it for MT, which is our default API type.


We can do this, and:

- Put all companies on a single `companies.md` page, with each company probably having a few bullet points mentioning what they offer. Instead of each company having it's own dedicated separate page. 

- Exclude companies from appearing in the search bar within the website to avoid a company to appear first instead of the product (if needed).

I think this will give us clearer and consistent slugs, and more scalability. Also, the URLs will have some difference and probably make it easier for SEO purposes, if someone is looking for an APE or QE.

(cc) @cefoo / @bittlingmayer

bittlingmayer:
Fine with me, if it's fine with @cefoo.

cefoo:
Thanks, makes sense, Tom!

> Regarding the location of the page, we can either keep it under `more/`, or bring it up front on the sidebar..?
> 
> What do you think?

It makes sense in the position it is right now, since More also has People, Associations, etc... But at the same time, I feel like having companies services listed on that page adds more value to it, other than just business info that might not be appealing to all, and that could be brought to the forefront.

IDK, @bittlingmayer?

PR #654, opened 2025-08-26
Display Models, QEs, APEs on home page

tovmasharrison:
# Description

Displays Models, QEs, APEs on the second row under APIs, Integrations and Languages

### Checklist:


tovmasharrison:
@cefoo Maybe we should rename APIs to Translation APIs, then, QE APIs and APE APIs?

bittlingmayer:
Sure, that works.

(translation APIs, quality estimation APIs, automatic post-editing APIs)

In fact, our current "APIs" is misleading because it doesn't include QE, APE no Routers ie Routing APIs which are also APIs.

(Maybe Integrations does include them though? @tovmasharrison can you check that and fix if not?)

I am also open to other words (systems, providers, engines...) or ways to totally sidestep this.

(Most folks get them via integrations anyway, not directly via APIs, but we used APIs draw a clear line between these systems and translation apps or features, that can't be used elsewhere.)

tovmasharrison:
@cefoo / @bittlingmayer 

- I swapped the location of models and integrations on the landing page grid
- I changed the naming for APIs to Translation APIs on the landing grid and across all other relevant areas.

bittlingmayer:
Works for me

